PR #5, opened 2018-06-18
Handle Sphinx changes related to `apidoc` location

jakirkham:
In Sphinx 1.7, it appears that `apidoc` was moved from `sphinx` to `sphinx.ext`. This adopts an `ImportError` based workaround to handle both cases. Also reverts a previous attempt to solve this issue, which did not end up working.

xref: https://github.com/sphinx-doc/sphinx/issues/5104

### DIFF
--- a/rtd_requirements.txt
+++ b/rtd_requirements.txt
@@ -1,0 +1,4 @@
+mock
+pillow
+sphinx
+sphinx_rtd_theme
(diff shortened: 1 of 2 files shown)


rtd_requirements.txt
```diff
@@ -0,0 +1,4 @@
+mock
+pillow
+sphinx
+sphinx_rtd_theme
```
jakirkham: Oops, this shouldn't be here. Dropping in PR ( https://github.com/dask/dask-image/pull/6 ).

Issue #7, opened 2018-06-26
Migrate work from dask-image org


jakirkham:
Would be good to consolidate the work from [the dask-image org libraries]( https://github.com/dask-image/ ) in this library. While it made sense to keep those libraries small and subject-oriented in the beginning to keep work focused and allow it progress incrementally, it now makes more sense to have one library that consolidates this functionality under a single API. Should improve usability and visibility of this work. Also should help by providing a singular point to engage users and developers alike.

jakirkham:
Working on this in PR ( https://github.com/dask/dask-image/pull/10 ).

PR #10, opened 2018-08-09
Migrate dask-image org's subpackages

jakirkham:
Fixes https://github.com/dask/dask-image/issues/7

Pulls in the subpackages from the [dask-image org]( https://github.com/dask-image ). Tries to preserve the git history here so that it can be revisited as needed. Planning on squashing it for `master`, but will keep a `historic` branch around so that it can be retrieved easily. Not going to attempt lots of cleanup or handle lots of review comments in this migration process in order to move quickly on this, but happy to raise issues for any problems identified to revisit at a later date.


jakirkham:
Merging to complete the migration. If anyone sees problems in here, please raise an issue. Thanks.

PR #36, opened 2018-08-30
Require SciPy pre-1.0.0

jakirkham:
It seems that SciPy 1.0.0+ causes some tests to fail. So go ahead and require SciPy to be pre-1.0.0 to ensure things don't break accidentally. Will need to revisit the cause of the failures separately and address them.


jakirkham:
> Will need to revisit the cause of the failures separately and address them.

Revisiting in PR ( https://github.com/dask/dask-image/pull/37 ).

PR #45, opened 2018-09-01
Breakout common changes in fourier_gaussian

jakirkham:
In all cases `sigma` is squared, the result is divided by `2`, and the negative of the result taken. These can be factored out into a common change to just `sigma` that is reused in all of the other computations. Admittedly Dask is smart enough to capture the reused computation of `sigma` squared, but it is not a symbolic algebra library. So it won't be able to pick up on the other changes, but we can. This should simplify the resulting graph and make it easy to reuse some common changes to `sigma` before `tensordot` is computed over multiple arrays and these changes applied.

### DIFF
--- a/dask_image/ndfourier/__init__.py
+++ b/dask_image/ndfourier/__init__.py
@@ -65,8 +65,9 @@ def fourier_gaussian(input, sigma, n=-1, axis=-1):
     )
 
     # Compute Fourier transformed Gaussian
+    scale = - (sigma ** 2) / 2
     gaussian = dask.array.exp(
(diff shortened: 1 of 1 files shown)


jakirkham:
Should add that if `sigma` is a concrete value, this computation occurs during graph construction. Thus Dask need not know anything about this computation. If `sigma` is a Dask Array, then Dask will construct nodes in the graph, but there should be significantly fewer nodes. Plus Dask should be able to fuse these `ufunc` operations into a single task.

On a more general note, as `sigma` is much smaller than the `gaussian` constructed. This computation should be much faster regardless of when it occurs.

PR #61, opened 2018-09-07
Use custom kernel for extrema

jakirkham:
Rewrites `extrema` to use its own custom kernel instead of relying on other functions to compute parts of the result.

### DIFF
--- a/dask_image/ndmeasure/__init__.py
+++ b/dask_image/ndmeasure/__init__.py
@@ -97,20 +97,45 @@ def extrema(input, labels=None, index=None):
         input, labels, index
     )
 
-    min_lbl = minimum(
-        input, labels, index
-    )
-    max_lbl = maximum(
-        input, labels, index
-    )
-    min_pos_lbl = minimum_position(
-        input, labels, index
-    )
-    max_pos_lbl = maximum_position(
-        input, labels, index
+    out_dtype = numpy.dtype([
+        ("min_val", input.dtype),
+        ("min_pos", numpy.int),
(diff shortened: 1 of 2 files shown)


dask_image/ndmeasure/__init__.py
```diff
@@ -97,20 +97,41 @@ def extrema(input, labels=None, index=None):
...
-        input, labels, index
+
+    extrema_lbl = {k: extrema_lbl[k] for k in extrema_lbl.dtype.names}
+
+    for pos_key in ["min_pos", "max_pos"]:
+        pos_1d = extrema_lbl[pos_key]
+        if not pos_1d.ndim:
+            pos_1d = pos_1d[None]
+
+        pos_nd = _utils._unravel_index(pos_1d, input.shape)
```
jakirkham: Note to self: This is essentially where the tests are failing. It occurs when indices has a N-D shape with `N > 1`. Not totally sure why that affects us here and not in `maximum_position` or `minimum_position`. It may have something to do with using a structured array type to return the result. Requires further inspection.

jakirkham: Addressed by PR ( https://github.com/dask/dask-image/pull/65 ) and PR ( https://github.com/dask/dask-image/pull/66 ).

jakirkham:
One of the tests failed due to a networking error. Unfortunately CircleCI is not letting me restart it from the GUI. To avoid eating up a bunch of time rerunning all CIs will merge and ignore this failure. Will keep an eye on `master` to see if this reoccurs.

PR #64, opened 2018-09-10
Bump dask-sphinx-theme to 1.0.4

jakirkham:


mrocklin:
FYI I would expect several more tiny version bumps in the next couple weeks.

jakirkham:
Makes sense. Should we worry about staying up-to-date or would it be better to wait until things settle a bit?

mrocklin:
If the readership of dask-image was very high and the flaws in
dask-sphinx-theme were egregious (some old versions had incorrect links)
then I would say lets stay up to date.  Currently I don't think that either
of those are true and I imagine that you're busy, so in your situation I
wouldn't bother.  That's obviously a choice for you to make though :)

jakirkham:
Ok, thanks Matt. I'll merge this one since it is here, but I'll let future ones slide until things settle or you tell me otherwise.

jakirkham:
Should add this is another effective method to deal with package caching issues. :wink:

PR #75, opened 2018-09-30
Rewrite `histogram` to use `labeled_comprehension`

jakirkham:
To simplify the `histogram` implementation a bit, rewrite it using `labeled_comprehension`. Make use of `functools.partial` to bind arguments to `histogram` so that it can be used with `labeled_comprehension`. Also simplify the internal `_histogram` function into a basic kernel function following these changes. Set the return type of `labeled_comprehension` as `object` to match that of what `histogram` used prior.


jakirkham:
This results in a Dask Array of `object` type being returned instead of a NumPy array of `object` type filled with Dask Delayed objects. Both will reach the same result when computed, but the former is easier to work with.

PR #85, opened 2018-10-10
Update HISTORY for 0.2.0 release  [ci skip]

jakirkham:


jakirkham:
Marked as 0.2.0 as the [return type]( https://github.com/dask/dask-image/pull/75#issuecomment-425702826 ) of `histogram` changed due to PR ( https://github.com/dask/dask-image/pull/75 ). Though expect that return type change largely goes unnoticed (or possibly is preferred) since most code will work just the same.

Issue #101, opened 2019-02-10
Coding style conventions/guide


jni:
We had some discussions on #94 about coding style. @jakirkham suggested we open a specific issue to discuss them.

I have three specific things that I find "inelegant" (yes, I know, it's imprecise) as currently implemented in `dask_image`.

1. fully specified imports, e.g. `dask.array` instead of `da`, `numpy` instead of `np`. Some of these abbreviations are very ingrained in the community and the full specification is surprising to read. (Not to mention annoying to write.)
2. Code definitions in `__init__.py`. Again, this is unconventional. `scikit-image` doesn't use these and still gets a nice API generation from sphinx, so I don't think it's a major hurdle to get sphinx to behave well for a more conventional code structure.
3. The import name with an underscore in it. ;)

I'll throw a bonus idea in here: we should rename `input` in all the ndimage functions. It was a bad choice from SciPy, and they're stuck with it, but I don't think dask-image should be bound by it.

Anyway, whatever is decided with the above points, it should probably be codified somewhere, and preferably it should reference the style guides for bigger projects, so that these aren't "just for dask-image" conventions.

GenevieveBuckley:
To document our decision today, we will use `image` and `label_image` as names for arrays passed into dask-image functions.

This is a move away from strict adherence to the scipy API (we don't like using the name `input` for arrays passed into functions because input is also the name of a python built in command). Choosing `image` and `label_image` mirrors scikit-image's decision, and they've put a fair bit of thought into this subject already.

PR #102, opened 2019-02-10
Optimize out first `where` in `label`

jakirkham:
This optimize out the first call to `where` in `label` as that case should be a no-op as it merely adds `0` to the first block's label image.

---

3. The pull request should work for Python 2.7, 3.5, 3.6, and 3.7. Check

### DIFF
--- a/dask_image/ndmeasure/__init__.py
+++ b/dask_image/ndmeasure/__init__.py
@@ -215,8 +215,9 @@ def label(input, structure=None):
     # First, label each block independently, incrementing the labels in that
     # block by the total number of labels from previous blocks. This way, each
     # block's labels are globally unique.
-    for index, cslice in zip(numpy.ndindex(*input.numblocks),
-                             dask.array.core.slices_from_chunks(input.chunks)):
+    block_iter = zip(numpy.ndindex(*input.numblocks),
+                     dask.array.core.slices_from_chunks(input.chunks))
+    for index, cslice in block_iter:
         input_block = input[cslice]
         labeled_block, n = _label.block_ndi_label_delayed(input_block,
                                                           structure)
--- a/dask_image/ndmeasure/__init__.py
+++ b/dask_image/ndmeasure/__init__.py
@@ -6,6 +6,7 @@ __email__ = "kirkhamj@janelia.hhmi.org"
 
 import collections
 import functools
+import itertools
 
 import numpy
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Thoughts @jni?

jni:
Can you benchmark this for some reasonable volume sizes? My view is that this makes the code significantly more complex (although it's a nice implementation! I didn't know about starmap!), with what in most cases will be a very minor performance benefit. (<1% improvement in a volume with 100 blocks.)

jakirkham:
> Can you benchmark this for some reasonable volume sizes?

Not sure I understand the purpose of a benchmark here. This removes a step (as opposed to changing it to an alternative implementation). As such it should only take less time. Could you please clarify what you are looking to learn from this?

> My view is that this makes the code significantly more complex...

Hmm...what about the current implementation do you find is more complex? What things do you think could be done to address this?

jni:
> As such it should only take less time. Could you please clarify what you are looking to learn from this?

**How much** less time? =)

> what about the current implementation do you find is more complex?

Honestly, I think I'm pretty good with dealing with iterators, but I find the starmap expression "work". Perhaps this optimisation can wait until we update to `arr.blocks`? Then there will be no need for itertools and lambdas.

jakirkham:
> ...I find the starmap expression "work".

Thanks for your feedback. My guess is that line was partially confusing as we are performing the selection after the zip (hence the use of starmap). Have rewritten it to perform the selection before the zip using map. What are your thoughts on this expression?

jni:
@jakirkham Maybe we should chat on #101 before going any further here... =P Specifically, I find the `_pycompat` stuff pretty awful... But, as I said last time, this is your library. ;)

jakirkham:
Unfortunately the change failed on Python 2 without them.

If the point is more a dislike of `_pycompat` generally, that seems orthogonal to this PR (especially given `_pycompat` proceeds it). If your concerns are style related, please do spell them out in that xref'd issue. If not, please feel free to raise a new issue.

At this point, my inclination is to merge. This PR is useful to us for the single chunk case. There may need to be a few more PRs for that case.

jni:
I'm just unclear about whether to `_pycompat` is better than `starmap`. Either one is acceptable, pick whichever is nicer for you.

Would a single-chunk-shortcut decorator be useful for the library?

jakirkham:
My guess was something like commit ( ac85bcd9c878890d496dd4d470b464326097cb5c ) would read better to you. Though I may not fully understand why the `starmap` expression was difficult to follow.

Not sure. Would probably need a better idea of what you have in mind? Maybe worth a new issue?

jni:
@jakirkham you're right, that commit is indeed nice, and I find the result very readable. Perhaps I found the `_pycompat` visual noise overwhelming.

jakirkham:
Will give issue ( https://github.com/dask/dask-image/issues/101 ) some thought and follow up there.

PR #108, opened 2019-03-11
Add dask-image quickstart documentation

GenevieveBuckley:
# Documentation update
Added:
* Brief text for quickstart.rst in the dask-image docs
* Links to the dask-examples repository (specifically of interest is dask-image-quickstart,ipynb)

Is this the right way to update the docs? I checked the restructured text formatting with http://rst.ninjs.org but should I be trying to run sphinx to check left hand sidebar updates to have a 'quickstart' tab too?

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -62,14 +62,6 @@ See the dask_array_documentation_ for more detail on general array operations.
 
     result = function_name(images)
 
-Remember that dask computations are lazy. 
-That's a great thing to have, most of the time.
-If you need to force dask to run the computation, use the `compute()` method.
-
-.. code-block:: python
-
-    result.compute()
-
 
 Further Reading
 ---------------
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -22,9 +22,9 @@ We highly recommend checking out the dask-image-quickstart.ipynb notebook
 
 https://github.com/dask/dask-examples
 
-All the example notebooks are available to launch with mybinder and test out
-interactively.
-
+You can find the dask-image quickstart notebook in the `applications` folder
+of this repository. All the example notebooks are available to launch with
+mybinder and test out interactively.
 
 
 An Even Quicker Start
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,0 +1,77 @@
+.. highlight:: shell
+
+==========
+Quickstart
+==========
+
+
+Importing dask-image
+--------------------
+Import dask image is with an underscore, like this example:
+
+.. code-block:: python
+
+    import dask_image.imread
+    import dask_image.ndfilters
+
+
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Calling dask-image functions is also easy.
+.. code-block:: python
+
+    import dask_image.ndfilters
+    blurred_image = dask_image.ndfilters.gaussian_filter(images, sigma=10)
+
+
+If you have a function that accepts input numpy arrays,
+it is also likely to accept dask arrays without any modification.
```
jakirkham: I'm not sure about this statement. There are a few cases this might run into issues (e.g. calls to `asarray` or some NumPy functions). Plus it's probably more general than what we'd like to discuss in dask-image (i.e. maybe a statement like this belongs in Dask's docs). Would it make sense to link out to Dask Array documentation for more general array operations or is there something in particular we'd like to say about image data here?

docs/quickstart.rst
```diff
@@ -0,0 +1,77 @@
...
+If you have a function that accepts input numpy arrays,
+it is also likely to accept dask arrays without any modification. 
+
+.. code-block:: python
+
+    result = my_function(images)
+
+Remember that dask computations are lazy. 
+That's a great thing to have, most of the time.
+If you need to force dask to run the computation, use the `compute()` method.
```
jakirkham: As this image data could be larger than memory, we should be somewhat careful recommending compute. Maybe it is worth noting some other options like `persist` and `store`. This is probably a situation where we can't go into too much detail, but we could have a brief intro followed by a link to appropriate Dask Array documented (e.g. maybe [this]( http://docs.dask.org/en/latest/array-creation.html#store-dask-arrays )). What do you think?

jakirkham:
To build the docs locally, just create [this environment]( https://github.com/dask/dask-image/blob/master/environment_doc.yml ), `cd` into the [`docs` directory]( https://github.com/dask/dask-image/tree/master/docs ), and run `make`.

docs/quickstart.rst
GenevieveBuckley: Yes, that would be the careful and sensible thing to do here. Changed & link added.

docs/quickstart.rst
GenevieveBuckley: Good point. I think I'll remove this bit altogether for now.

GenevieveBuckley:
> To build the docs locally, just create [this environment](https://github.com/dask/dask-image/blob/master/environment_doc.yml), `cd` into the [`docs` directory](https://github.com/dask/dask-image/tree/master/docs), and run `make`.

Too easy! The built html looks good to me.

jakirkham:
Thanks @GenevieveBuckley. Looks great! 😄

PR #112, opened 2019-04-01
threshold_local function for dask-image ndfilters

GenevieveBuckley:
The scikit-image local threshold methods return a numpy array the same size as the input image data. For very large input data this could become problematic. This PR allows threshold_local() to return a dask array instead.

Reference: [skimage.filters.threshold_local()](https://github.com/scikit-image/scikit-image/blob/8f733a4cd7cd5ababbd7d0aea3277bd48270a87c/skimage/filters/thresholding.py#L140) code.

Things to note:

1. The tests for the local gaussian threshold method seem to suggest the results of the dask-image wrapped scipy.ndimage.gaussian_filter() are different to the straight scipy.ndimage.gaussian_filter(). Both use the 'reflect' mode, both have the same keyword arguments. Am I missing something here?
1. I've implemented the local mean threshold method is implemented differently than in scikit-image, purely because I didn't see why it shouldn't be n-dimensional.
1. I haven't checked that the block_size argument contains only even valued numbers, because I don't think we need to?


GenevieveBuckley:
flake8 is failing because there is a very long URL in a docstring. Do you have a suggestion for how to best shorten or ignore it? (Google hasn't been especially helpful here)

`/home/circleci/test/dask_image/ndfilters/_thresholding.py:58:80: E501 line too long (109 > 79 characters)`

GenevieveBuckley:
Note to self: double check how we're handling the chunk size for sigma, I think this won't work so well if we pass in multidimensional values for `block_size`

```python
sigma = (da.from_array(block_size, chunks=1) - 1) / 6.0
```

GenevieveBuckley:
Ok, I think this is ready for review @jakirkham 

The failing travis builds seem unrelated, xcode 6.4 is now deprecated on Travis. I'm tinkering with trying to fix that here: https://github.com/dask/dask-image/pull/113

GenevieveBuckley:
Ok tests are all green so it's ready for review @jakirkham

martindurant:
@jakirkham , seems like a decent amount of work here, do you intend to look into it?

GenevieveBuckley:
Any chance you might take a quick look at this @jni? Also happy to hear thoughts from @martindurant if he wants to weigh in.

I'm considering merging within the next week (ahead of PyConAU on Friday Sept 4th, or at the very least ahead of SciPy Japan in October) instead of letting this languish. It might need a little work to get running with the GPU, but that should be doable.

martindurant:
@dask/maintenance , who is the best person currently for dask-image maintenance?

mrocklin:
> @dask/maintenance , who is the best person currently for dask-image maintenance?

Heh, my usual answer to this is @GenevieveBuckley :)

GenevieveBuckley:
Sorry to confuse you @martindurant. I'm able to merge PRs here (and am planning to do that for this one soon-ish).

It's just in general I'm reluctant to merge my own PRs (especially for new features) without any input from other people. But for a project with only two maintainers, that may not always be practical!

PR #122, opened 2019-08-05
Add simple test for center_of_mass

Mause:

3. The pull request should work for Python 3.5, 3.6, and 3.7. Check

### DIFF
--- a/.appveyor_support/environments/tst_py37.yml
+++ b/.appveyor_support/environments/tst_py37.yml
@@ -9,7 +9,7 @@ dependencies:
   - wheel==0.32.3
   - coverage==4.5.2
   - flake8==3.7.2
-  - pytest==4.2.0
+  - pytest==5.2.0
   - pytest-flake8==1.0.4
   - dask==1.1.1
   - numpy==1.15.4
--- a/.circleci/environments/tst_py37.yml
+++ b/.circleci/environments/tst_py37.yml
@@ -9,7 +9,7 @@ dependencies:
   - wheel==0.32.3
   - coverage==4.5.2
   - flake8==3.7.2
-  - pytest==4.2.0
+  - pytest==5.2.0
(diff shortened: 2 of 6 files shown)


GenevieveBuckley:
Ah, I see why you were not sure if the issue contained a legitimate bug.

So, we only see the bug if an array containing *integer values* is passed to `dask_image.ndmeasure.center_of_mass()`. The array you make in this test has float values in it, since `numpy.random.random()` [outputs arrays of float values](https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.random.random.html).

For this test to show the bug, it would need to go something like this:
```python
a = np.random.random(shape) * 255 # multiply so values are in 8-bit range 0-255
a = a.astype(np.uint8) . # convert to unsigned 8-bit integer data
d = da.from_array(a, chunks=chunks)
```

This also means that the actual fix for the issue should also be a single line, converting the array chunks to float type before running the computation in `dask_image.ndmeasure.center_of_mass()`

GenevieveBuckley:
I think the fix should happen in [this line](https://github.com/dask/dask-image/blob/7e8932be51a4de44ed83000f061fbcd4fe9668f4/dask_image/ndmeasure/__init__.py#L108) of  `dask-image/dask_image/ndmeasure/__init__.py`:

```python
out_dtype = numpy.dtype([("com", image.dtype, (image.ndim,))])
```
`image.dtype` needs to be changed to always be float type (instead of just reflecting the type of the input array like it does now). That's *probably* going to look like this instead:

```python
out_dtype = numpy.dtype([("com", float, (image.ndim,))])
```

Mause:
I figured the issue was the data I was testing with, but wasn't sure how to fix it - I'll do that now

GenevieveBuckley:
Thanks @Mause!

PR #145, opened 2020-06-08
Use tifffile for testing instead of scikit-image

jakirkham:
As scikit-image has dropped their vendored copy of tifffile ( https://github.com/scikit-image/scikit-image/pull/4235 ), switch to using tifffile directly for testing our ability to read TIFFs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ test_requirements = [
     "flake8 >=3.4.1",
     "pytest >=3.0.5",
     "pytest-flake8 >=0.8.1",
-    "scikit-image >=0.12.3",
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
Out of interest, how do you choose which version of `tifffile` (or any dependency) to pin to? Is is just whichever version you happen to have in your environment right now that's working well, i.e. `dependency>=the-version-I-have-now`?

jakirkham:
Honestly it was sort of arbitrary. Older versions may work fine as well. They switched from semantic version to date versioning here. So I picked that. Happy to change that if there’s a different version we expect.

Issue #129, opened 2019-10-15
Release 0.3?


stuarteberg:
Looks like there have been various bugfixes and improvements since the last release.  Perhaps the most interesting one is the new distributed implementation of `ndmeasure.label()` from @jni and @jakirkham (#94).

Is it time for a new release?  I suppose you'll want to bump straight to 0.3, since Python 2.7 has been dropped (#119).

GenevieveBuckley:
It might be, @jakirkham what's your feeling? I'd be happy to learn the release process and make notes to add to the docs.

Maybe we can sneak https://github.com/dask/dask-image/pull/112 into this one too

jni:
Ping @jakirkham 

Also feel free to grant @GenevieveBuckley and I (and @stuarteberg!) access rights for PyPI if, as I suspect, you are too busy with NVIDIA-y things to continue maintaining this!

leifdenby:
I would be happy to help with this too. `dask_image.ndmeasure.area` has been added since `0.2` too which is really great(!)

If the versioning is following semver would changing the argument `labels`  to `label_image` in the `dask_image.ndmeasure` functions be considered a breaking a change and warrant making the next version `1.0.0`? Sorry if I've missed discussion on this. This library has saved me a lot of time, so I'd like to contribute back if I can

GenevieveBuckley:
This still needs followup, the bottleneck here is the ability to upload to PyPI. 

If you're more comfortable with it @jakirkham, we could use [Github secrets](https://github.com/dask/dask-image/settings/secrets) to tie those credentials to this repository, and then use github actions to automate the build/release. I don't have access to the [conda feedstock](https://github.com/conda-forge/dask-image-feedstock), but I think that should just update itself automatically after a successful PyPI release?

I'm happy to do the legwork on [setting up a github action for this](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/). Thanks to Kira, napari have just set up the same thing, so we could use that as a model here: https://github.com/napari/napari/blob/master/.github/workflows/make_release.yml

jni:
@leifdenby 

> considered a breaking a change and warrant making the next version 1.0.0

Note that in semver, everything is fair game in the 0.x series. =) Typically, packages in the SciPy stack have allowed breaking changes between minor versions in early 0.x releases (with or without warnings, depending on the maturity of the library).

jakirkham:
[...]

PR #146, opened 2020-06-08
Update release dependencies

GenevieveBuckley:
As discussed [here](https://github.com/dask/dask-image/issues/129#issuecomment-640380667), I think it's a good idea for us to include a minimal set of dependencies to allow users to open most image files with our `imread` function. I want things "just work" out of the box.

The extra dependencies proposed here are:
* `tifffile` (for tiff files)
* `imageio` (for most other files)
* `imageio-ffmpeg` (is also required by the pims imageio reader)
* `jpype` (for bioformats files)

### DIFF
--- a/.appveyor_support/environments/tst_py36.yml
+++ b/.appveyor_support/environments/tst_py36.yml
@@ -14,7 +14,7 @@ dependencies:
   - dask==0.16.1
   - numpy==1.11.3
   - scipy==1.1.0
-  - scikit-image==0.14.2
+  - tifffile==2018.10.18
   - pims==0.4.1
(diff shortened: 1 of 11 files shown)


GenevieveBuckley:
Closing this. 

As discussed, maybe in the future we could depend directly on `tifffile` and `imageio` (if we don't also need to include `imageio-ffmpeg`, see [here for details](https://github.com/soft-matter/pims/pull/358)).

Also a  good idea - adding a note to our `imread` docs describing how users can install `jpype` themselves for bioformats (but `dask-image` will not automatically include `jpype` as a dependency)

PR #148, opened 2020-06-24
Match template

hungyiwu:
Hi, I ported `skimage.feature.match_template` here and worked out some hiccups. I added a test that shows it generates the same result as skimage, copied over the doc string from skimage and added a few lines at the top about the key differences, and passed flake8 and pytest.

I cound't find where to add this in README.rst and https://travis-ci.org/dask/dask-image/pull_requests gave me 404. How can I make this right?

===

3. The pull request should work for Python 3.6, 3.7, and 3.8. Check


dask_image/feature/template.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _window_sum_3d(image, window_shape):
+
+    window_sum = _window_sum_2d(image, window_shape)
+
+    window_sum = da.cumsum(window_sum, axis=2)
+    window_sum = (window_sum[:, :, window_shape[2]:-1]
+                  - window_sum[:, :, :-window_shape[2] - 1])
+
+    return window_sum
```
jni, line 27: I think this can be made n-dimensional without too much difficulty...

```python
integral_image = image
for ax in range(image.ndim):
    integral_image = da.cumsum(integral_image, axis=ax)
window_sums = integral_image
for ax in range(image.ndim):
    start = [slice(None),] * image.ndim
    end = [slice(None),] * image.ndim
    start[ax] = slice(0, -window_shape[ax])
    end[ax] = slice(window_shape[ax], None)
    # (not sure why the -1 margins were necessary, btw?)
    window_sums = window_sums[end] - window_sums[start]
```

Not sure if that's exactly right but should be close...!

hungyiwu: Ah thank you for the suggestion! I'll work on this soon

hungyiwu:
I realized there's no need to port this function; I thought the memory issue I encountered was due to not having this function ported, but turns out I was using `dask.array` the wrong way 🤦

Initially I wanted to do this:
```
img = tifffile.imread('/path/to/img.tif')
response = skimage.feature.match_template(img, ...)
```
This image is ~8 GB on disk and this operation crashed on my laptop (16 GB RAM)

Then I did this:
```
img = dask_image.imread.imread('/path/to/img.tif')
response = skimage.feature.match_template(img, ...).compute()
```
This still crashed, and I thought porting `skimage.feature.match_template` can solve this

So I ported the function and did this:
```
img = dask_image.imread.imread('/path/to/img.tif')
response = dask_image.feature.match_template(img, ...).compute()
```
This code worked with a larger portion of the image, but still crashed for the whole image

Then I started profiling this code using `dask.diagnostics.{Profiler, ResourceProfiler}` and found out the following code works for the whole image on my laptop:
```
img = dask_image.imread.imread('/path/to/img.tif')
dask.array.to_hdf5('/path/to/img.hdf5', '/arr1', img)
```
and then:  
```
f = h5py.File('/path/to/img.hdf5', 'r')['arr1']
x = da.from_array(f, chunk_size=...)
match_fn = lambda block: skimage.feature.match_template(block, ..., pad_input=False)
response = x.map_overlap(match_fn, ..., trim=False, chunks=x.chunks)
da.to_hdf5('/path/to/output.hdf5', '/response1', response)
```

Lessons learned:
* `dask_image.imread.imread('image.tif')` took more time and memory than converting `.tif` to `.hdf5` and then `dask.array.from_array(h5py.File('image.hdf5')['dataset_name'])` combined  
* `dask.array.compute(...); dask.array.to_hdf5(...)` means first dumping the result into RAM before saving to disk, while running `dask.array.to_hdf5(...)` directly handles memory footprint better  
* `func(dask.array)` is not the same as `dask.array.map_blocks(func)`; initially I thought they are the same under the hood

If this makes sense to you than I'll just close this PR. Any thoughts?

hungyiwu:
I'm closing this as `skimage.feature.match_template` can be used with `dask.array.map_blocks` and there's no need to port this feature

Issue #133, opened 2020-01-06
GPU Support


jakirkham:
Would be useful to support these operations on GPUs as well. One way to do this might be to dispatch using CuPy.

jakirkham:
Here's what convolution would look like if someone rolled their own today using CuPy 7.0.0 and Dask 2.9.1. There are a few rough edges, but we could improve on this over time both in dask-image and upstream.

```python
import numpy
import numpy as np

import cupy
import cupy as cp

import cupyx
import cupyx.scipy
import cupyx.scipy.ndimage
import cupyx.scipy.ndimage.filters

import dask
import dask.array
import dask.array as da

import dask_image
import dask_image.ndfilters
import dask_image.ndfilters._utils

s = (100, 110)
a = da.from_array(cp.arange(int(np.prod(s)), dtype=cp.float32).reshape(s), chunks=10)
w = cp.ones(a.ndim * (3,), dtype=cp.float32)

origin = dask_image.ndfilters._utils._get_origin(w.shape, 0)
depth = dask_image.ndfilters._utils._get_depth(w.shape, origin)
depth, boundary = dask_image.ndfilters._utils._get_depth_boundary(a.ndim, depth, "none")



r = a.map_overlap(cupyx.scipy.ndimage.filters.convolve,
                  depth=depth,
                  boundary=boundary,
                  dtype=a.dtype,
                  weights=w,
                  mode="reflect",
                  cval=0.0,
                  origin=origin)
r._meta = cp.empty(r.ndim * (0,), dtype=r.dtype)

r.compute()
```

GenevieveBuckley:
Linking to this blog post from Matt about mixing dask and cupy arrays, since I found it helpful to read and it's relevant for anyone who comes across this issue: https://matthewrocklin.com/blog/work/2019/01/03/dask-array-gpus-first-steps

jakirkham:
Also we may be able to use `__array_function__` to dispatch over different operations (like `convolve` here). Some discussion along these lines in issue ( https://github.com/scipy/scipy/issues/10204 ).

GenevieveBuckley:
I haven't had a chance to look much into cupyimg (BSD-3 licensed), but that could also be useful to look at here: https://github.com/mritools/cupyimg

jakirkham:
Yeah I’ve seen that. There’s also some work to add equivalent functions to CuPy.

[...]

PR #150, opened 2020-08-03
GPU support proposal

GenevieveBuckley:
Here's the code showing how we might implement GPU support for the `convolve` function. Relevant issue: https://github.com/dask/dask-image/issues/133

Two questions:
1. If we use this pattern, how should I update the docstring wrapper? (Do I just decorate the dispatcher instance, or a method thereof?)
2. Is there a free tier including GPU support available to us for CI? Ideally we'd want to have tests for cupy arrays too.

You can try this PR out like this:
```
pip uninstall dask_image
pip install git+https://github.com/GenevieveBuckley/dask-image.git@dispatch
```

```python
import numpy as np
import cupy as cp
import dask.array as da
from dask_image.ndfilters import convolve

s = (10, 10)

# convolve with numpy
a = da.from_array(np.arange(int(np.prod(s)), dtype=np.float32).reshape(s), chunks=5)
w = np.ones(a.ndim * (3,), dtype=np.float32)
result = convolve(a, w)
result.compute()

# convolve with cupy
a = da.from_array(cp.arange(int(np.prod(s)), dtype=cp.float32).reshape(s), chunks=5)
w = cp.ones(a.ndim * (3,), dtype=cp.float32)
result = convolve(a, w)
result.compute()
```

### DIFF
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -18,7 +18,7 @@ def convolve(image,
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
     result = image.map_overlap(
-        dispatch_convolve,
+        dispatch_convolve(image),
         depth=depth,
         boundary=boundary,
         dtype=image.dtype,
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -4,7 +4,7 @@ import numpy as np
 import scipy.ndimage.filters
 
 from . import _utils
-from ..utils import Dispatcher
+from ..utils._dispatcher import Dispatcher
 
 convolve = Dispatcher(name="convolve")
 
@@ -46,7 +46,9 @@ def register_cupy():
 
     @convolve.register(cupy.ndarray)
     def cupy_convolve(*args, **kwargs):
-        return convolve_func(cupyx.scipy.ndimage.filters.convolve, *args, **kwargs)
+        return convolve_func(cupyx.scipy.ndimage.filters.convolve,
+                             *args,
+                             **kwargs)
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.correlate)
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -1,9 +1,8 @@
 # -*- coding: utf-8 -*-
-
 import scipy.ndimage.filters
 
 from . import _utils
-from ..utils._dispatcher import convolve_dispatch
+from ..utils._dispatcher import dispatch_convolve, check_arraytypes_compatible
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.convolve)
@@ -12,15 +11,18 @@ def convolve(image,
              mode='reflect',
              cval=0.0,
              origin=0):
+    check_arraytypes_compatible(image, weights)
+
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
     result = image.map_overlap(
-        convolve_dispatch(image),
+        dispatch_convolve,
         depth=depth,
         boundary=boundary,
         dtype=image.dtype,
+        meta=image._meta,
         weights=weights,
         mode=mode,
         cval=cval,
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -4,14 +4,18 @@
 import scipy.ndimage.filters
 
 from . import _utils
+from ..utils import Dispatcher
+
+convolve = Dispatcher(name="convolve")
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.convolve)
-def convolve(image,
-             weights,
-             mode='reflect',
-             cval=0.0,
-             origin=0):
+def convolve_func(func,
+                  image,
+                  weights,
+                  mode='reflect',
+                  cval=0.0,
+                  origin=0):
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
@@ -28,6 +32,20 @@ def convolve(image,
     )
 
     return result
+
+
+@convolve.register(np.ndarray)
+def numpy_convolve(*args, **kwargs):
+    return convolve_func(scipy.ndimage.filters.convolve, *args, **kwargs)
+
+
+@sizeof.register_lazy("cupy")
+def register_cupy():
+    import cupy
+
+    @convolve.register(cupy.ndarray)
+    def cupy_convolve(*args, **kwargs):
+        return convolve_func(cupyx.scipy.ndimage.filters.convolve, *args, **kwargs)
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.correlate)
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -1,27 +1,23 @@
 # -*- coding: utf-8 -*-
 
-import numpy as np
 import scipy.ndimage.filters
 
 from . import _utils
-from ..utils._dispatcher import Dispatcher
-
-convolve = Dispatcher(name="convolve")
+from ..utils._dispatcher import convolve_dispatch
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.convolve)
-def convolve_func(func,
-                  image,
-                  weights,
-                  mode='reflect',
-                  cval=0.0,
-                  origin=0):
+def convolve(image,
+             weights,
+             mode='reflect',
+             cval=0.0,
+             origin=0):
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
     result = image.map_overlap(
-        func,
+        convolve_dispatch(image),
         depth=depth,
         boundary=boundary,
         dtype=image.dtype,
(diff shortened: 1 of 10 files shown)


GenevieveBuckley:
Tests for python 3.6 are failing because the way we are looking up if it's a numpy or cupy dask array did not exist in dask version 0.16.1

I'll try and figure out when that was introduced.

GenevieveBuckley:
The coveralls report is a very good argument for separating out the dispatcher function registrations from the rest of the code. Your 100% coverage is a great motivator to keep things in good shape (I confess I do need that pressure/motivation!)

dask_image/ndfilters/_conv.py
```diff
@@ -1,23 +1,27 @@
...
+                  mode='reflect',
+                  cval=0.0,
+                  origin=0):
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
     result = image.map_overlap(
-        scipy.ndimage.filters.convolve,
+        func,
```
jakirkham, line 24: Did you also try using the `convolve` `Dispatch` object here? I'm curious if that would simplify things.

GenevieveBuckley: I didn't think to try that. It might simplify things, I'll give it a go.

dask_image/ndfilters/_conv.py
```diff
@@ -17,7 +17,7 @@ def convolve(image,
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
     result = image.map_overlap(
-        scipy.ndimage.filters.convolve,
+        convolve_dispatch(image),
```
GenevieveBuckley, line 20: This is a little bit weird because we need to pass an argument to the dispatcher (so it can check the array type) but it returns a bare function

dask_image/ndfilters/_conv.py
```diff
@@ -17,7 +17,7 @@ def convolve(image,
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
```
GenevieveBuckley, line 17: I haven't yet tested what happens if you pass in mismatched dask array types (eg: a cupy-backed image dask array, and a numpy-backed weights dask array). There might need to be a check & informative error message added here for that

dask_image/ndfilters/_conv.py
GenevieveBuckley: It does seem a bit simpler overall

dask_image/utils/_dispatcher.py
```diff
@@ -0,0 +1,32 @@
...
+        meth = self.dispatch(type(arg._meta))
+        return meth(arg, *args, **kwargs)
+
+
+convolve_dispatch = Dispatcher(name="convolve_dispatch")
+
+
+@convolve_dispatch.register(np.ndarray)
+def numpy_convolve(*args, **kwargs):
+    return scipy.ndimage.filters.convolve
```
jakirkham, line 22:
```suggestion
    return scipy.ndimage.filters.convolve(*args, **kwargs)
```

dask_image/utils/_dispatcher.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@convolve_dispatch.register_lazy("cupy")
+def register_cupy():
+    import cupy
+    import cupyx.scipy.ndimage
+
+    @convolve_dispatch.register(cupy.ndarray)
+    def cupy_convolve(*args, **kwargs):
+        return cupyx.scipy.ndimage.filters.convolve
```
jakirkham, line 32:
```suggestion
        return cupyx.scipy.ndimage.filters.convolve(*args, **kwargs)
```

dask_image/utils/_dispatcher.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Dispatcher(Dispatch):
+    """Simple single dispatch for different dask array types."""
+
+    def __call__(self, arg, *args, **kwargs):
+        """
+        Call the corresponding method based on type of dask array.
+        """
+        meth = self.dispatch(type(arg._meta))
+        return meth(arg, *args, **kwargs)
```
jakirkham, line 14: I think with the changes below we can drop this.

dask_image/ndfilters/_conv.py
jakirkham: That's a good point. Made a couple suggestions below that may help. Though please let me know if you already tried them and ran into issue.

dask_image/utils/_dispatcher.py
GenevieveBuckley: Is this really what we want here? I had thought the bare function itself was what we wanted returned, and that's what gets put in as the first argument to `map_overlap`. I didn't try it this way, though.

dask_image/ndfilters/_conv.py
jakirkham: We may also need to make this change as well. Sorry for missing that before.

```suggestion
        convolve_dispatch,
```

dask_image/utils/_dispatcher.py
jakirkham: That's typically how it is used like [with `sizeof`]( https://github.com/dask/dask/blob/72de42cdde88ba01b4d762147e4105892a6535a2/dask/sizeof.py#L17-L22 ).

As `Dispatch` instances (like `convolve_dispatch`) should behave like a function, I think this should work here. Though please let me know if that is not the case.

dask_image/utils/_dispatcher.py
GenevieveBuckley: I'm not sure I understand what you mean by dropping this.
The important part here is that we use `type(arg._meta)` to dispatch according to the type of arrays contained in the dask array. So that part matters, but other than that I don't mind what we do.

dask_image/utils/_dispatcher.py
GenevieveBuckley: I think that's because `sizeof` isn't using `map_overlap` or similar, so you need to pass the arguments into the function for it to work. Whereas here we have `image.map_overlap(func, *args, *kwargs)` and I am trying to replace func with the dispatch object (which should return `func` as a bare function object that is ingested into `map_overlap`).

I might just be confused about what you're trying to say here though

dask_image/ndfilters/_conv.py
GenevieveBuckley: > We may also need to make this change as well. Sorry for missing that before.

This makes more sense to me now, thanks!

GenevieveBuckley:
I've only just noticed this, but it seems that when we pass in cupy/dask arrays and dispatch to cupyx.scipy what we get back is a dask array backed by numpy (not cupy). This seems unintuitive to me, is that what's supposed to happen here? I'm not exactly sure at which point of the computation things get turned back into numpy arrays.

Here's a gist that shows what I'm describing: https://gist.github.com/GenevieveBuckley/332a50020ce35baf3cea763e95d12bd8

Did you see anything like this when you did your initial exploration @jakirkham?

GenevieveBuckley:
Oh hang on, I see this did happen to you and you've added a line `r._meta = cp.empty(r.ndim * (0,), dtype=r.dtype)` in your [code here](https://github.com/dask/dask-image/issues/133)  to overwrite what we see there. Further digging reveals the chunks in our output dask array are actually cupy (so that's a relief).

Maybe the best way to handle this overall is to add a chunktype kwarg to `map_blocks`, similar to the `dtype` kwarg? We could handle it internally the way you do in your script within dask-image (if you didn't want to pin to a much higher minimum dask version, or wait on a new dask release if we made a PR there)

GenevieveBuckley:
Update: there *is* a `meta` kwarg you can use in `map_overlap`, but if I use `meta=image._meta` then that seems to break our dispatching (throws up a `TypeError: No dispatch for <class 'cupy.core.core.ndarray'>`)

GenevieveBuckley:
Updates:
1. The issue with `meta=` is now fixed in the dask master branch, courtesy of Peter https://github.com/dask/dask/pull/6494
2. I'm having some trouble with cupy + register_lazy for dispatch. I haven't distilled this down to a minimal example yet though, still not entirely sure where the problem is creeping in.

jakirkham:
cc @grlee77 (who may be interested in this as well 😉)

jakirkham:
Also thanks for the updates here Genevieve. Sorry for the slow reply. Glad Peter's work was helpful. Happy to work through CuPy dispatch issues (even non-minimal ones) if you need another set of eyes 🙂

GenevieveBuckley:
Thanks @jakirkham I appreciate your willingness to help track down the weirdness. (Apologies for my own slow response, I was on leave from work last week. Now I'm back, I'd love to use the upcoming conference as motivation to get this in and done by the end of August if possible - although I understand if your time is much more limited here).

Here is a non-minimal example of the problem with lazy cupy dispatching. You will need a python environment with:
* dask=2.23.0 (the latest release)
* numpy
* cupy

Then you can run the code example [from my first comment in this thread](https://github.com/dask/dask-image/pull/150#issue-461908070) with dask-image installed from each of these two branches in turn:
```
pip install git+https://github.com/GenevieveBuckley/dask-image.git@working-cupy-non-lazy-dispatch
```
This branch works! (But it *requires* cupy to be installed by the user, which is very inconvenient)

```
pip install git+https://github.com/GenevieveBuckley/dask-image.git@failing-cupy-lazy-dispatch
```
This branch does not work :(

There's only a single commit difference between these branches, where I turn the lazy cupy dispatch into a non-lazy dispatch.

GenevieveBuckley:
Ok, I understand what's going on now. Because `map_overlap` delays computation, the lazy cupy dispatch function isn't registered at all until the first time you try and run `compute()` - by which time it's too late to actually use it for the computation. If you try the example above, and run `result.compute()` *twice* in a row, then the second time it works. (You can also take a look at `dispatch_convolve._lookup` before and after each compute call to see this directly)

We didn't see this problem earlier because before I had `image.map_overlap(dispatch_convolve(image), ...)` which executes `dispatch_convolve` and registers the cupy function straight away.

For this reason I think we need to return to the pattern we had earlier, where the dispatcher returns a bare function object. (I am open to other solutions if you wanted to suggest something different). Once you're happy with it, we can roll it out to the rest of the `dask-image` library.

GenevieveBuckley:
Speaking of rolling it out, I've started another branch here https://github.com/dask/dask-image/pull/151 to make those broader changes to the rest of the library. Both ndfourier and ndmeasure have some numpy specific array creation code hiding in there, so if you have any suggestions for handling those areas I'd love to hear your thoughts.

GenevieveBuckley:
This has served its purpose as a concrete example to discuss the implementation of https://github.com/dask/dask-image/issues/133. So, I'll close this PR now in favour of https://github.com/dask/dask-image/pull/151 and future work.

PR #151, opened 2020-08-17
GPU support for ndfilters & imread modules

GenevieveBuckley:
This branch contains broader support for GPU array computation with cupy & dask-image.

Related PR: https://github.com/dask/dask-image/pull/150
Related issue: https://github.com/dask/dask-image/issues/133

TODO
- [x] ndfilters - done! (Note: @jakirkham used a clever way to avoid repetition with the `minimum`, `median`, and `maximum` filters, but this PR opts for a small amount of repetition for the sake of simplicity - it seemed easier to keep the docstring wrapper function and the dispatching working well this way)
- [x] imread - done!
- [x] Tests explicitly using cupy arrays (even if we have to mark them as skipped on CI)

For future PRs
* GPU support for ndmorph - easy to implement ([see gist here](https://gist.github.com/GenevieveBuckley/ecb81a13c93139ae8dc9cc1e50f1e2ca)), but it relies on having cupy support for the functions `binary_dilation()` and `binary_erosion()` which does not currently exist.
* GPU support for ndfourier - problem: lots of numpy specific array creation in `dask_image.ndfourier._utils`
* GPU support for ndmeasure - problem: lots of numpy specific array creation in distributed labelling. I still need to look at what needs to be done to support all the ndmeasure functions.


dask_image/imread/__init__.py
```diff
@@ -43,6 +45,12 @@ def imread(fname, nframes=1):
     if (nframes != -1) and not (nframes > 0):
         raise ValueError("`nframes` must be greater than zero.")
 
+    if arraytype == "numpy":
+        arrayfunc = numpy.asanyarray
+    elif arraytype == "cupy":
+        import cupy
+        arrayfunc = cupy.asanyarray
```
GenevieveBuckley, line 52: I'm sure there's a nicer way to handle this instead of an if clause, this could be refactored (now or later) if someone has a better idea for it.

dask_image/imread/__init__.py
```diff
@@ -43,6 +45,12 @@ def imread(fname, nframes=1):
...
+
     with pims.open(fname) as imgs:
         shape = (len(imgs),) + imgs.frame_shape
         dtype = numpy.dtype(imgs.pixel_type)
```
GenevieveBuckley, line 56: FYI: we don't need to change the line `numpy.dtype(imgs.pixel_type)`, this works fine as is.

GenevieveBuckley:
FYI, I'm going to tidy this up and merge it within the next week so I can talk about it at PyConAU (Friday Sept 4th).

Planning to:
* Add some brief tests for dask cupy arrays, with a skipif statement so pytest will skip them on our CI and only run them if cupy is available
* Tidy up: use a proper dispatch mechanism in the `imread` and add a check that the relevant library (eg: cupy) is actually installed and working on the user's system
* not worry about GPU support for the `ndfourier` and `ndmeasure` modules now, we can do that later
* not worry about the slight drop in test coverage, some of this is unavoidable if we can't run tests with cupy on our CI

jakirkham:
Sorry for the slow reply. Have been occupied with the RAPIDS release, which is wrapping up this week. Would add that I'll be out next week. Though no need to hold things up because of that.

Should have some time tomorrow and Friday to look over things. So will follow up after having read over this more recent work.

GenevieveBuckley:
That's totally fine, no need to apologize!

While I *like* having your opinion, I by no means *expect* it. Mostly I announce what I'm doing to give you the opportunity to raise objections (about implementation details, or about whether a new feature should be included or scrapped).

jakirkham:
Of course. Just trying to give you some insight into what is going on 😉 This remains important 🙂

GenevieveBuckley:
### Request for help: lazy dispatch for `correlate`.
I'm at my wits end here, so if you have any suggestions @jakirkham I'd be happy to hear them.

What works:
* Dispatch for `convolve` with lazy cupy function registration
    * The implementation for `convolve` is virtually identical to the implementation for `correlate`, I don't understand why they are behaving differently.
    * All the other ndfilter function dispatchers seem to work fine too, it's just `correlate` that's not working
* The non-lazy version of the `correlate` dispatch for cupy arrays also works. You can try this yourself by commenting [lines 66-67](https://github.com/GenevieveBuckley/dask-image/blob/32d4ac8e9512a27ccd5ce75e6d80893235bd2f63/dask_image/dispatch/_dispatch_ndfilters.py#L66-L67) in `dask-image/dask_image/dispatch/_dispatch_ndfilters.py` and then indenting [lines 68-73](https://github.com/GenevieveBuckley/dask-image/blob/32d4ac8e9512a27ccd5ce75e6d80893235bd2f63/dask_image/dispatch/_dispatch_ndfilters.py#L68-L73) below that.

What doesn't work:
* Dispatch for `correlate` with lazy cupy function registration

chrisroat:
Regarding binary_dilation and binary_erosion for ndmorph, there is cupy support through the following repo.  Not sure if you can borrow or depend on it:
https://github.com/mritools/cupyimg

GenevieveBuckley:
> 
> 
> Regarding binary_dilation and binary_erosion for ndmorph, there is cupy support through the following repo. Not sure if you can borrow or depend on it:
> https://github.com/mritools/cupyimg

Good to know, thanks @chrisroat

GenevieveBuckley:
> 
> 
> Regarding binary_dilation and binary_erosion for ndmorph, there is cupy support through the following repo. Not sure if you can borrow or depend on it:
> https://github.com/mritools/cupyimg

Good to know, thanks @chrisroat

PR #158, opened 2020-09-01
Add GPU support for local_threshold with method='mean'

GenevieveBuckley:
Adds the fix suggested in https://github.com/cupy/cupy/issues/3909


GenevieveBuckley:
Worth noting: it takes a moment or two for the CUDA kernel to compile when you run `compute()` for the first time. Subsequent times don't need compilation and so things are fast. If this is a problem for people, we could look at triggering compilation of the CUDA kernel *at the time we build the dask task graph* instead (ie: in the dispatch function right after we define it).

Issue #24, opened 2018-08-30
SciPy N-D interpolation functions


jakirkham:
_From @jakirkham on May 16, 2017 18:35_

The list of functions below are what that this library intends to support with Dask Arrays. These all come from SciPy. In particular they come from [`scipy.ndimage.interpolation`]( https://docs.scipy.org/doc/scipy-0.19.0/reference/ndimage.html#interpolation ). The intent is to emulate their behavior with Dask Arrays as wrapping them is not likely to work. Will also include tests to verify these retain the behavior that the SciPy functions ordinarily have.

* [ ]  [affine_transform]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.affine_transform.html )
* [ ]  [geometric_transform]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.geometric_transform.html )
* [ ]  [map_coordinates]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.map_coordinates.html )
* [ ]  [rotate]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.rotate.html )
* [ ]  [shift]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.shift.html )
* [ ]  [spline_filter]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.spline_filter.html )
* [ ]  [spline_filter1d]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.spline_filter1d.html )
* [ ]  [zoom]( https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.ndimage.zoom.html )

_Copied from original issue: dask-image/dask-ndinterp#3_

wtbarnes:
@jakirkham I'm very interested in implementations of `affine_transform` and `shift`. I'd be happy to put some effort into implementing both of these. Have there been any efforts already to implement either of these functions?

GenevieveBuckley:
> @jakirkham I'm very interested in implementations of `affine_transform` and `shift`. I'd be happy to put some effort into implementing both of these. Have there been any efforts already to implement either of these functions?

[...]

PR #117, opened 2019-07-13
Rename the input argument to image in the ndimage functions

akhalighi:
This PR implements the following:

1. The wrappers for functions ported from `scipy.ndimage` are modified to rename the `input` argument to `image`. 

2. The `_get_docstring` utility function for `filters` is modified to:
     - replace the `input` keyword with `image` when porting docstrings.
     - drop `Examples` from the original docstring in `scipy.ndimage` since they most probably will be invalid for the `dask-image` wrappers.

Contributes to #101

PR #159, opened 2020-10-09
Added dask implementation of ndimage.affine_transform in new ndinterp directory.

m-albert:
This PR aims to add `dask-image` support for `scipy.ndimage.affine_transform` (https://github.com/dask/dask-image/issues/24).

Code changes:
  - added new directory for ndinterp
  - included `affine_transform` implementation in `__init__.py`
  - added test that compares output to scipy for some test cases

Differences to `ndimage.affine_transformation`:
  - currently, prefiltering is not supported (changing the output in case of interpolation `order > 1`)
  - default order is 1
  - currently, matrix is expected to have shape (ndim,) or (ndim, ndim) (because offset needs to be set for each chunk)
  - modes 'reflect', 'mirror' and 'wrap' may yield different results

This implementation includes suggestions by @grlee77 about how to provide the correct spline order footprint.

Also, as @grlee77 pointed out, dask leveraged prefiltering could be added as it consists of preprocessing the input array with `ndimage.spline_filter`, which is a separable filter that requires the full array along a given axis (see https://github.com/dask/dask-image/issues/24#issuecomment-706165593). However, even if implemented it might make sense to prevent this prefiltering from being executed by default, as the chunking it requires would (in my understanding) lead to the input array being computed entirely (or at least in parts that are much larger than the relevant input blocks).

### DIFF
--- a/dask_image/dispatch/_dispatch_ndinterp.py
+++ b/dask_image/dispatch/_dispatch_ndinterp.py
@@ -20,11 +20,11 @@ def numpy_affine_transform(*args, **kwargs):
 
 
 @dispatch_affine_transform.register_lazy("cupy")
-def register_cupy_convolve():
+def register_cupy_affine_transform():
     import cupy
     import cupyx.scipy.ndimage
 
-    @dispatch_convolve.register(cupy.ndarray)
+    @dispatch_affine_transform.register(cupy.ndarray)
     def cupy_affine_transform(*args, **kwargs):
 
         return cupyx.scipy.ndimage.affine_transform
--- a/dask_image/dispatch/_dispatch_ndinterp.py
+++ b/dask_image/dispatch/_dispatch_ndinterp.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+import numpy as np
+from scipy import ndimage
+
+from ._dispatcher import Dispatcher
+
+__all__ = [
+    "dispatch_affine_transform",
+    "dispatch_asarray",
+]
+
+
+dispatch_affine_transform = Dispatcher(name="dispatch_affine_transform")
+
+# ================== affine_transform ==================
+@dispatch_affine_transform.register(np.ndarray)
+def numpy_affine_transform(*args, **kwargs):
+    return ndimage.affine_transform
+
+
+@dispatch_affine_transform.register_lazy("cupy")
+def register_cupy_convolve():
+    import cupy
+    import cupyx.scipy.ndimage
+
+    @dispatch_convolve.register(cupy.ndarray)
+    def cupy_affine_transform(*args, **kwargs):
+
+        return cupyx.scipy.ndimage.affine_transform
+
+
+dispatch_asarray = Dispatcher(name="dispatch_asarray")
+
+# ================== affine_transform ==================
+@dispatch_asarray.register(np.ndarray)
+def numpy_asarray(*args, **kwargs):
+    return np.asarray
+
+
+@dispatch_asarray.register_lazy("cupy")
+def register_cupy_asarray():
+    import cupy
+
+    @dispatch_asarray.register(cupy.ndarray)
+    def cupy_asarray(*args, **kwargs):
+
+        return cupy.asarray
--- a/dask_image/dispatch/_dispatch_ndinterp.py
+++ b/dask_image/dispatch/_dispatch_ndinterp.py
@@ -32,7 +32,7 @@ def register_cupy_affine_transform():
 
 dispatch_asarray = Dispatcher(name="dispatch_asarray")
 
-# ================== affine_transform ==================
+# ===================== asarray ========================
 @dispatch_asarray.register(np.ndarray)
 def numpy_asarray(*args, **kwargs):
     return np.asarray
--- a/dask_image/ndinterp/__init__.py
+++ b/dask_image/ndinterp/__init__.py
@@ -129,8 +129,11 @@ def resample_chunk(chunk, image, matrix, offset, order, func_kwargs, block_info=
 
         # for some reason the behaviour becomes inconsistent with ndimage
         # if leaving out the -1 in the next line
-        rel_image_i[dim] = np.floor(rel_image_i[dim]) - order // 2 - 1
+        rel_image_i[dim] = np.floor(rel_image_i[dim]) - order // 2
         rel_image_f[dim] = np.floor(rel_image_f[dim]) - order // 2 + order
+
+        if order == 0:
+            rel_image_i[dim] -= 1
 
     # clip image coordinates to image extent
     for dim, s in zip(range(n), image_shape):
--- a/dask_image/ndinterp/__init__.py
+++ b/dask_image/ndinterp/__init__.py
@@ -3,7 +3,7 @@
 
 import numpy as np
 import dask.array as da
-from scipy import ndimage
+from scipy.ndimage import affine_transform as ndimage_affine_transform
 import warnings
 
 
@@ -72,8 +72,35 @@ def affine_transform(
     if output_chunks is None:
         output_chunks = image.shape
 
-    # process kwargs
+    # Perform test run to ensure parameter validity.
+    ndimage_affine_transform(np.zeros([0] * image.ndim),
+                             matrix,
+                             offset)
 
+    # Make sure parameters contained in matrix and offset
+    # are not overlapping, i.e. that the offset is valid as
+    # it needs to be modified for each chunk.
+    # Further parameter checks are performed directly by
+    # `ndimage.affine_transform`.
+
+    matrix = np.asarray(matrix)
+    offset = np.asarray(offset).squeeze()
+
+    # these lines were copied and adapted from `ndimage.affine_transform`
+    if (matrix.ndim == 2 and matrix.shape[1] == image.ndim + 1 and
+            (matrix.shape[0] in [image.ndim, image.ndim + 1])):
+        # if matrix.shape[0] == image.ndim + 1:
+        #     exptd = [0] * image.ndim + [1]
+        #     if not np.all(matrix[image.ndim] == exptd):
+        #         msg = ('Expected homogeneous transformation matrix with '
+        #                'shape %s for image shape %s, but bottom row was '
+        #                'not equal to %s' % (matrix.shape, image.shape, exptd))
+        #         raise ValueError(msg)
+        # assume input is homogeneous coordinate transformation matrix
+        offset = matrix[:image.ndim, image.ndim]
+        matrix = matrix[:image.ndim, :image.ndim]
+
+    # process kwargs
     # prefilter is not yet supported
     if 'prefilter' in kwargs:
         if kwargs['prefilter'] and order > 1:
@@ -107,12 +134,18 @@ def resample_chunk(chunk, image, matrix, offset, order, func_kwargs, block_info=
     image_shape = image.shape
     chunk_shape = chunk.shape
 
+    # extract coordinates of chunk edges
     chunk_offset = [i[0] for i in block_info[0]['array-location']]
-
     chunk_edges = np.array([i for i in np.ndindex(tuple([2] * n))])\
         * np.array(chunk_shape) + np.array(chunk_offset)
 
-    rel_image_edges = np.dot(matrix, chunk_edges.T).T + offset
+    # map output chunk edges onto input image coordinates
+    # to define the input region relevant for the current chunk
+    if matrix.ndim == 1 and len(matrix) == image.ndim:
+        rel_image_edges = matrix * chunk_edges + offset
+    else:
+        rel_image_edges = np.dot(matrix, chunk_edges.T).T + offset
+
     rel_image_i = np.min(rel_image_edges, 0)
     rel_image_f = np.max(rel_image_edges, 0)
 
@@ -157,7 +190,7 @@ def resample_chunk(chunk, image, matrix, offset, order, func_kwargs, block_info=
     # o' = o + Mx0 - y0
     offset_prime = offset + np.dot(matrix, chunk_offset) - rel_image_i
 
-    chunk = ndimage.affine_transform(rel_image,
+    chunk = ndimage_affine_transform(rel_image,
                                      matrix,
                                      offset_prime,
                                      output_shape=chunk_shape,
--- a/dask_image/ndinterp/__init__.py
+++ b/dask_image/ndinterp/__init__.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+
+
+import numpy as np
+import dask.array as da
+from scipy import ndimage
+
+
+__all__ = [
+    "affine_transform",
+]
+
+
+def affine_transform(
+        input,
+        matrix,
+        offset=0.0,
+        output_shape=None,
+        output_chunks=None,
+        **kwargs
+):
+    """Apply an affine transform using Dask.
+    `ndimage.affine_transformation` for Dask Arrays. For every
+    output chunk, only the slice containing the relevant part
+    of the input is passed on to `ndimage.affine_transformation`.
+
+    Notes
+    -----
+        Differences to `ndimage.affine_transformation`:
+        - currently, prefiltering is not supported
+        - currently, matrix is expected to have shape (ndim,) or (ndim, ndim)
+          (because offset needs to be set for each chunk)
+        - modes 'reflect', 'mirror' and 'wrap' may yield different results
+
+        To do:
+        - optionally use cupyx.scipy.ndimage.affine_transform
+
+        Arguments equal to `ndimage.affine_transformation`,
+        except for `output_chunks`.
+
+    Parameters
+    ----------
+    input : array_like (Numpy Array, Dask Array, ...)
+        The input array.
+    matrix : array (ndim,) or (ndim, ndim)
+        Transformation matrix.
+    offset : array (ndim,)
+        Transformation offset.
+    output_shape : array (ndim,), optional
+        The size of the array to be returned.
+    output_chunks : array (ndim,), optional
+        The chunks of the output Dask Array.
+
+    Returns
+    -------
+    affine_transform : Dask Array
+        A dask array representing the transformed output
+
+    """
+
+    def resample_chunk(chunk, _matrix, _offset, _order, func_kwargs, block_info=None):
+
+        n = chunk.ndim
+        input_shape = input.shape
+        chunk_shape = chunk.shape
+
+        chunk_offset = [i[0] for i in block_info[0]['array-location']]
+
+        chunk_edges = np.array([i for i in np.ndindex(tuple([2] * n))])\
+            * np.array(chunk_shape) + np.array(chunk_offset)
+
+        rel_input_edges = np.dot(_matrix, chunk_edges.T).T + _offset
+        rel_input_i = np.min(rel_input_edges, 0)
+        rel_input_f = np.max(rel_input_edges, 0)
+
+        # expand relevant input to contain footprint of the spline kernel
+        # according to
+        #
+        # https://github.com/scipy/scipy/blob/
+        # 9c0d08d7d11fc33311a96d2ac3ad73c8f6e3df00/
+        # scipy/ndimage/src/ni_interpolation.c#L412-L419
+        #
+        # Also see https://github.com/dask/dask-image/
+        # issues/24#issuecomment-706165593
+        for dim in range(n):
+            if _order % 1:
+                start_lower = np.floor(rel_input_i[dim]) - _order // 2
+                start_upper = np.floor(rel_input_f[dim]) - _order // 2
+            else:
+                start_lower = np.floor(rel_input_i[dim] + 0.5) - _order // 2
+                start_upper = np.floor(rel_input_f[dim] + 0.5) - _order // 2
+
+            stop_upper = start_upper + _order
+            # for some reason the behaviour is different to ndimage's
+            # if leaving out the -1 in the next two lines
+            rel_input_i[dim] = start_lower - 1
+            rel_input_f[dim] = stop_upper + 1
+
+        # clip input coordinates to input image extent
+        for dim, upper in zip(range(n), input_shape):
+            rel_input_i[dim] = np.clip(rel_input_i[dim], 0, upper)
+            rel_input_f[dim] = np.clip(rel_input_f[dim], 0, upper)
+
+        rel_input_slice = tuple([slice(int(rel_input_i[dim]),
+                                       int(rel_input_f[dim]))
+                                 for dim in range(n)])
+
+        rel_input = input[rel_input_slice]
+
+        # modify offset to point into cropped input
+        # y = Mx + o
+        # coordinate substitution:
+        # y' = y - y0(min_coord_px)
+        # x' = x - x0(chunk_offset)
+        # then:
+        # y' = Mx' + o + Mx0 - y0
+        # M' = M
+        # o' = o + Mx0 - y0
+        offset_prime = _offset + np.dot(_matrix, chunk_offset) - rel_input_i
+
+        chunk = ndimage.affine_transform(rel_input,
+                                         _matrix,
+                                         offset_prime,
+                                         output_shape=chunk_shape,
+                                         order=_order,
+                                         prefilter=False,
+                                         **func_kwargs)
+
+        return chunk
+
+    if output_shape is None:
+        output_shape = input.shape
+
+    # process kwargs
+
+    # set default order to 1, warn if different
+    if 'order' in kwargs:
+        order = kwargs['order']
+        del kwargs['order']
+        if order > 1:
+            UserWarning('Currently, for order > 1 the output of'
+                        '`dask_image.ndinterp.affine_transform` can differ'
+                        'from that of `ndimage.affine_transform`')
+    else:
+        order = 1
+
+    # prefilter is not yet supported
+    if 'prefilter' in kwargs:
+        if kwargs['prefilter'] and order > 1:
+            UserWarning('Currently, `dask_image.ndinterp.affine_transform`'
+                        'does not support `prefilter=True`')
+        del kwargs['prefilter']
+
+    transformed = da.zeros(output_shape,
+                           dtype=input.dtype,
+                           chunks=output_chunks)
+
+    transformed = transformed.map_blocks(resample_chunk,
+                                         dtype=input.dtype,
+                                         _matrix=matrix,
+                                         _offset=offset,
+                                         _order=order,
+                                         func_kwargs=kwargs,
+                                         )
+
+    return transformed
--- a/dask_image/ndinterp/__init__.py
+++ b/dask_image/ndinterp/__init__.py
@@ -28,6 +28,8 @@ def affine_transform(
     -----
         Differences to `ndimage.affine_transformation`:
         - currently, prefiltering is not supported
+          (affecting the output in case of interpolation `order > 1`)
+        - default order is 1
         - currently, matrix is expected to have shape (ndim,) or (ndim, ndim)
           (because offset needs to be set for each chunk)
         - modes 'reflect', 'mirror' and 'wrap' may yield different results
@@ -58,76 +60,6 @@ def affine_transform(
 
     """
 
-    def resample_chunk(chunk, _matrix, _offset, _order, func_kwargs, block_info=None):
-
-        n = chunk.ndim
-        input_shape = input.shape
-        chunk_shape = chunk.shape
-
-        chunk_offset = [i[0] for i in block_info[0]['array-location']]
-
-        chunk_edges = np.array([i for i in np.ndindex(tuple([2] * n))])\
-            * np.array(chunk_shape) + np.array(chunk_offset)
-
-        rel_input_edges = np.dot(_matrix, chunk_edges.T).T + _offset
-        rel_input_i = np.min(rel_input_edges, 0)
-        rel_input_f = np.max(rel_input_edges, 0)
-
-        # expand relevant input to contain footprint of the spline kernel
-        # according to
-        #
-        # https://github.com/scipy/scipy/blob/
-        # 9c0d08d7d11fc33311a96d2ac3ad73c8f6e3df00/
-        # scipy/ndimage/src/ni_interpolation.c#L412-L419
-        #
-        # Also see https://github.com/dask/dask-image/
-        # issues/24#issuecomment-706165593
-        for dim in range(n):
-            if _order % 1:
-                start_lower = np.floor(rel_input_i[dim]) - _order // 2
-                start_upper = np.floor(rel_input_f[dim]) - _order // 2
-            else:
-                start_lower = np.floor(rel_input_i[dim] + 0.5) - _order // 2
-                start_upper = np.floor(rel_input_f[dim] + 0.5) - _order // 2
-
-            stop_upper = start_upper + _order
-            # for some reason the behaviour is different to ndimage's
-            # if leaving out the -1 in the next two lines
-            rel_input_i[dim] = start_lower - 1
-            rel_input_f[dim] = stop_upper + 1
-
-        # clip input coordinates to input image extent
-        for dim, upper in zip(range(n), input_shape):
-            rel_input_i[dim] = np.clip(rel_input_i[dim], 0, upper)
-            rel_input_f[dim] = np.clip(rel_input_f[dim], 0, upper)
-
-        rel_input_slice = tuple([slice(int(rel_input_i[dim]),
-                                       int(rel_input_f[dim]))
-                                 for dim in range(n)])
-
-        rel_input = input[rel_input_slice]
-
-        # modify offset to point into cropped input
-        # y = Mx + o
-        # coordinate substitution:
-        # y' = y - y0(min_coord_px)
-        # x' = x - x0(chunk_offset)
-        # then:
-        # y' = Mx' + o + Mx0 - y0
-        # M' = M
-        # o' = o + Mx0 - y0
-        offset_prime = _offset + np.dot(_matrix, chunk_offset) - rel_input_i
-
-        chunk = ndimage.affine_transform(rel_input,
-                                         _matrix,
-                                         offset_prime,
-                                         output_shape=chunk_shape,
-                                         order=_order,
-                                         prefilter=False,
-                                         **func_kwargs)
-
-        return chunk
-
     if output_shape is None:
         output_shape = input.shape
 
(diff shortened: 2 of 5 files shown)


dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,166 @@
...
+        The chunks of the output Dask Array.
+
+    Returns
+    -------
+    affine_transform : Dask Array
+        A dask array representing the transformed output
+
+    """
+
+    def resample_chunk(chunk, _matrix, _offset, _order, func_kwargs, block_info=None):
```
jakirkham, line 61: Would recommend making this a separate function as opposed to a closure. It's cheaper to pickle a function in a module than a closure. Also it is smaller in pickled form, which makes sending it to Dask workers less costly.

m-albert: That makes sense!

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+from scipy import ndimage
+
+
+__all__ = [
+    "affine_transform",
+]
+
+
+def affine_transform(
+        input,
```
GenevieveBuckley, line 15: One slight difference between dask-image and scipy is that we use `image` instead of `input`
https://github.com/dask/dask-image/pull/117

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        output_shape = input.shape
+
+    # process kwargs
+
+    # set default order to 1, warn if different
+    if 'order' in kwargs:
+        order = kwargs['order']
+        del kwargs['order']
+        if order > 1:
+            UserWarning('Currently, for order > 1 the output of'
```
GenevieveBuckley, line 73: This is a very vague user warning, what actually happens here if order > 1 and how is it different that what we might expect?

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # define (random) transformation
+    matrix = np.eye(n) + (np.random.random((n, n)) - 0.5) / 5.
+    offset = (np.random.random(n) - 0.5) / 5. * np.array(im.shape)
+
+    # define resampling options
+    output_shape = [int(a / 2)] * n
+    output_chunks = [16] * n
+
+    from scipy import ndimage
```
GenevieveBuckley, line 42: Shouldn't this import go at the top with the others? Or did you have a special reason for putting it down here instead?

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,58 @@
...
+    Notes
+    -----
+        Currently, prefilter is disabled and therefore the output
+        of `dask_image.ndinterp.affine_transformation` is compared
+        to `prefilter=False`.
+    """
+
+    # define test image
+    a = 25
+    im = np.random.random([a] * n)
```
GenevieveBuckley, line 28: Suggestion purely for readability & consistency throughout dask-image: let's use `image` instead of the abbreviation `im`.

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+        * np.array(chunk_shape) + np.array(chunk_offset)
+
+    rel_input_edges = np.dot(matrix, chunk_edges.T).T + offset
+    rel_input_i = np.min(rel_input_edges, 0)
+    rel_input_f = np.max(rel_input_edges, 0)
+
+    # expand relevant input to contain footprint of the spline kernel
+    # according to
+    #
+    # https://github.com/scipy/scipy/blob/
```
GenevieveBuckley, line 121: I'm happy for comment lines with long urls on them to extend past 80 characters, and would prefer it since it's hard to copy-paste a url that runs over multiple lines. I think you'd just need to tell flake8 to ignore it by adding a ` # noqa: E501` comment to the end of the line.

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,58 @@
...
+# -*- coding: utf-8 -*-
+
+import pytest
+
+import dask_image.ndinterp as da_ndinterp
+
+import numpy as np
+import dask.array as da
+
+
```
GenevieveBuckley, line 11: Are you able to add a couple more test cases to cover the currently untested lines of your function?
https://coveralls.io/builds/34077648/source?filename=dask_image/ndinterp/__init__.py

Things like:
* no output shape specified
* no order specified
* if order is not an integer value
* Do we get the user warnings we expect for prefiltering, etc. if we pass in those keyword arguments, using [pytest warns](https://docs.pytest.org/en/3.0.3/recwarn.html)

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+    # 9c0d08d7d11fc33311a96d2ac3ad73c8f6e3df00/
+    # scipy/ndimage/src/ni_interpolation.c#L412-L419
+    #
+    # Also see https://github.com/dask/dask-image/
+    # issues/24#issuecomment-706165593
+    for dim in range(n):
+        if order % 1:
```
GenevieveBuckley, line 128: Can you leave a short (in-line?) comment for future programmers describing why we have this modulo checking here.

GenevieveBuckley:
What do we think, is `interp` the right name for this new subpackage @jakirkham ?

jakirkham:
Thus far the names have been based off of the submodules that SciPy used to collect functionality from `ndimage`. If we continue to follow that model, then this would be `interpolation` based off of `scipy.ndimage.interpolation`. Though we don't necessarily need to follow that naming scheme.

dask_image/ndinterp/__init__.py
m-albert: That's better, changed!

dask_image/ndinterp/__init__.py
m-albert: You're right, added: "Specifically, no spline prefiltering is performed and therefore the output can contain more blur."

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
m-albert: Hmm no good reason, put it at the top with the others as you suggested.

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
m-albert: Done

dask_image/ndinterp/__init__.py
m-albert: That makes sense, didn't know this trick :)

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
m-albert: Good point! Actually, `output_shape=None` had lead to an exception. That's fixed and more tests have been added. Let me know if you think some more would make sense.

dask_image/ndinterp/__init__.py
m-albert: Good that you mentioned this. I took over the code @grlee77 posted here https://github.com/dask/dask-image/issues/24#issuecomment-706165593, but when comparing it to the [c code](https://github.com/scipy/scipy/blob/9c0d08d7d11fc33311a96d2ac3ad73c8f6e3df00/scipy/ndimage/src/ni_interpolation.c#L412-L419%20#%20noqa:%20E501) I think there might be some problems with it. I changed it to follow the c code now, however as commented I'm still unsure about a `-1` which is required to pass tests (that I extended to contain more edge cases).

m-albert:
> What do we think, is interp the right name for this new subpackage @jakirkham ?

> Thus far the names have been based off of the submodules that SciPy used to collect functionality from ndimage. If we continue to follow that model, then this would be interpolation based off of scipy.ndimage.interpolation. Though we don't necessarily need to follow that naming scheme.

Didn't mean to impose anything :) I had taken the name for the new submodule from the archived https://github.com/dask-image/dask-ndinterp.

Related to that, I realised that for consistency with `scipy.ndimage`, `dask-image.ndmorph` could be renamed to `dask-image.ndmorphology`.

jakirkham:
Ha! Fair 😄 I think I disliked the deviation from the `nd*` theme so remedied that. Though again I don't think we need to be tied to that name either 😉

grlee77:
Performance here seems pretty good. I modified the test case use shape (256, 256, 256) for both input & output with chunk sizes of (48, 48, 48) and on a 10-core Skylake X CPU:

order | acceleration factor
------|--------------------
0 | 2.4
1 | 3.5
3 | 6.8
5 | 7.7

The acceleration factors can become a little better in some cases if the chunk size is tuned independently for each order, but this is pretty representative.

dask_image/ndinterp/__init__.py
grlee77: It does seem that which cases fail if the `- 1` is removed is not consistent from run to run of the tests (because the tests do not fix a random seed, different random offsets are tested each time).

However, I have only seen cases where `interp_order = 0` fail. All other orders always seem to pass without the ` - 1`.

grlee77 (commented):
Eventually, I think we want to parametrize over the `mode` argument as well, but I think it is okay to hold off on that now as the boundary handling has some issues upstream in SciPy that are being resolved in scipy/scipy#12767 and scipy/scipy#12776 (additional reviewers welcome for those PRs!).

Regarding support for `prefilter`: It is not strictly necessary to rechunk as I suggested previously. You can keep the same chunks as here, but with a larger boundary because even though the filters are IIR, the contributions from distant samples decays somewhat rapidly (how rapidly varies with spline order and depends on the magnitude of the filter poles). How large that boundary needs to be depends on the level of error that can be tolerated. It may be substantial (> 10 samples) to get error below 1e-7 as need to pass the default `numpy.allclose` comparison, but 5-10 samples might be enough for many practical applications. I am fine with adding prefilter support in a future PR, though.

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    of the image is passed on to `ndimage.affine_transformation`.
+
+    Notes
+    -----
+        Differences to `ndimage.affine_transformation`:
+        - currently, prefiltering is not supported
+          (affecting the output in case of interpolation `order > 1`)
+        - default order is 1
+        - currently, matrix is expected to have shape (ndim,) or (ndim, ndim)
+          (because offset needs to be set for each chunk)
```
jni, line 36: But you can split up the matrix _internally_ to the function, as is done in `scipy.ndimage`, right? So you could still support (ndim+1, ndim+1), then, as soon as you enter the function, you rip it up into an (ndim, ndim) matrix and (ndim,) offsets.

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,58 @@
...
+                                          prefilter=False)
+
+    # transform with dask-image
+    im_t_dask = da_ndinterp.affine_transform(dim, matrix, offset,
+                                             output_shape=output_shape,
+                                             output_chunks=output_chunks,
+                                             order=interp_order)
+    im_t_dask_computed = im_t_dask.compute()
+
+    print(im_t_scipy, im_t_dask_computed)
```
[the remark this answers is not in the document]
m-albert: Yes, thanks!

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,58 @@
...
+import pytest
+
+import dask_image.ndinterp as da_ndinterp
+
+import numpy as np
+import dask.array as da
+
+
+@pytest.mark.parametrize("n", [1, 2, 3, 4])
+@pytest.mark.parametrize("interp_order", [0, 1, 3])
```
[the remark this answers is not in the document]
m-albert: Ok! After adding more parameters to be iterated over I wasn't sure whether it'd be worth it to test all orders, but on my laptop tests take ~4s now, which I guess is still acceptable.

m-albert:
> Eventually, I think we want to parametrize over the mode argument as well, but I think it is okay to hold off on that now as the boundary handling has some issues upstream in SciPy that are being resolved in scipy/scipy#12767 and scipy/scipy#12776 (additional reviewers welcome for those PRs!).

That's a good point. Actually, in the current implementation only the modes `nearest` and `constant` would be supported in general. While it might be difficult to fully support the mode `wrap`, `reflect` and `mirror` could be more easily adapted by increasing the input regions as needed. Do you think it makes sense to implement this or will most users not need it?

> Regarding support for prefilter: It is not strictly necessary to rechunk as I suggested previously. You can keep the same chunks as here, but with a larger boundary because even though the filters are IIR, the contributions from distant samples decays somewhat rapidly (how rapidly varies with spline order and depends on the magnitude of the filter poles). How large that boundary needs to be depends on the level of error that can be tolerated. It may be substantial (> 10 samples) to get error below 1e-7 as need to pass the default numpy.allclose comparison, but 5-10 samples might be enough for many practical applications. I am fine with adding prefilter support in a future PR, though.

Interesting that 5-10 additional samples could already be enough to get prefiltering close enough. It might make sense to incorporate `prefilter` functionality in a future PR once `spline_filter` support is implemented in `dask-image`.

dask_image/ndinterp/__init__.py
m-albert: Yes, the parameters could be processed in the function. I guess at first I didn't want to copy this `ndimage` code that could be subject to changes:

    matrix = numpy.asarray(matrix, dtype=numpy.float64)
    if matrix.ndim not in [1, 2] or matrix.shape[0] < 1:
        raise RuntimeError('no proper affine matrix provided')
    if (matrix.ndim == 2 and matrix.shape[1] == input.ndim + 1 and
            (matrix.shape[0] in [input.ndim, input.ndim + 1])):
        if matrix.shape[0] == input.ndim + 1:
            exptd = [0] * input.ndim + [1]
            if not numpy.all(matrix[input.ndim] == exptd):
                msg = ('Expected homogeneous transformation matrix with '
                       'shape %s for image shape %s, but bottom row was '
                       'not equal to %s' % (matrix.shape, input.shape, exptd))
                raise ValueError(msg)
        # assume input is homogeneous coordinate transformation matrix
        offset = matrix[:input.ndim, input.ndim]
        matrix = matrix[:input.ndim, :input.ndim]
    if matrix.shape[0] != input.ndim:
        raise RuntimeError('affine matrix has wrong number of rows')
    if matrix.ndim == 2 and matrix.shape[1] != output.ndim:
        raise RuntimeError('affine matrix has wrong number of columns')
    if not matrix.flags.contiguous:
        matrix = matrix.copy()

But I agree with you that this is an important part of the function's interface and not being flexible with the parameter format could represent an obstacle to some use cases / users.

It would be great if the above code in `ndimage` was outsourced into a different function, to be reused here. But in the meantime it probably makes sense to just copy it here.

m-albert:
> Ha! Fair 😄 I think I disliked the deviation from the nd* theme so remedied that. Though again I don't think we need to be tied to that name either 😉

I agree consistency is nice! The question here seems to be which consistency to prioritise 😁.

@jakirkham @GenevieveBuckley Just let me know which name it should have :)

dask_image/ndinterp/__init__.py
jni: > I guess at first I didn't want to copy this ndimage code that could be subject to changes:

I literally wrote that exact code so that hurt. 🤣🤣🤣

> in the meantime it probably makes sense to just copy it here.

I agree. 😝 

> It would be great if the above code in ndimage was outsourced into a different function, to be reused here.

I think SciPy would be reluctant to add such a function to their public API anyway. So it would be a private function which we probably would want to copy!

m-albert: > I literally wrote that exact code so that hurt. 🤣🤣🤣

Haha my statement referred to potentially duplicating it rather than the code itself 😁😂👍.

>I think SciPy would be reluctant to add such a function to their public API anyway. So it would be a private function which we probably would want to copy!

You're right, no need for a new function.

I think I found a good solution now (see recent commit). At first, an `ndimage.affine_transform` test run on a dummy array checks that parameters are valid (using your code!). In a second step, only the two relevant parameter formats that overwrite the offset are handled. This should give full support for all the formats and avoids testing the duplicated code :)

grlee77:
> That's a good point. Actually, in the current implementation only the modes nearest and constant would be supported in general, right?. While it might be difficult to fully support the mode wrap, reflect and mirror could be more easily adapted by increasing the input regions if needed (with the risk of enlarging them significantly though). In your experience, do you think users would need this?

On second thought, given that we are just relying on SciPy internally to process the blocks, we do not need to wait for any of the boundary handling fixes upstream before testing all modes. The behavior should just match upstream SciPy. Once things are fixed in SciPy they will also automatically be fixed for dask-image as well.

I think `wrap` is the only case that will be problematic for this implementation because the values to be wrapped will not belong to the same chunk. 'reflect' and 'mirror' just reflect or mirror values that are already in the same chunk so I don't think they are a problem (you have already expanded the boundary enough to cover the filter footprint for the non-edge chunks).  I would just add something to explicitly raise an exception for `wrap`:

```Python
if mode == 'wrap':
    raise NotImplementedError(f"mode 'wrap' is not currently supported.")
```

I tried running the test locally with parameterized mode over ['constant', 'nearest', 'reflect', 'wrap', 'mirror'] and the only cases that failed were for 'wrap', so that seems to confirm the above.

m-albert:
> I think wrap is the only case that will be problematic for this implementation because the values to be wrapped will not belong to the same chunk. 'reflect' and 'mirror' just reflect or mirror values that are already in the same chunk so I don't think they are a problem (you have already expanded the boundary enough to cover the filter footprint for the non-edge chunks)

I think problems occur for pixels further away from the boundary than there are valid pixels in the given chunk. And even for sensible choices of chunking sizes this can occur quickly.

> I tried running the test locally with parameterized mode over ['constant', 'nearest', 'reflect', 'wrap', 'mirror'] and the only cases that failed were for 'wrap', so that seems to confirm the above.

For example, this test with just the identity transform fails:

```python
@pytest.mark.parametrize("n",
                         [1])
@pytest.mark.parametrize("interp_mode",
                         ['mirror'])
@pytest.mark.parametrize("input_output_shape_per_dim",
                         [(2, 3)])
@pytest.mark.parametrize("input_output_chunksize_per_dim",
                         [(2, 2)])
@pytest.mark.parametrize("matrix",
                         [np.eye(1)])
@pytest.mark.parametrize("offset",
                         [[0]])
```
with `ndimage.affine_transform` and `da_ndinterp.affine_transform` giving [0.5488135 , 0.71518937, 0.5488135 ] and [0.5488135 , 0.71518937, 0.71518937] respectively. The reason is that in the latter case, the last output chunk only bases its array extension on the second pixel.

So depending on the combinations of image shapes and chunk sizes, differences of this kind can easily occur. Before looking into this with more detail I thought problems would start to occur only far away from the boundary, but actually not even the first pixel is safe.

So we'd almost need to exclude all modes but 'constant' and 'nearest'.

jakirkham:
I wonder if we can add some dispatching logic based on the array type. That way we can run with NumPy or CuPy. FWIW Genevieve did some work on this for existing functions ( https://github.com/dask/dask-image/pull/151 ) ( https://github.com/dask/dask-image/pull/158 ). Should be pretty straightforward now that we have a framework for dispatching 🙂

m-albert:
> I wonder if we can add some dispatching logic based on the array type. That way we can run with NumPy or CuPy. FWIW Genevieve did some work on this for existing functions ( #151 ) ( #158 ). Should be pretty straightforward now that we have a framework for dispatching 🙂

@jakirkham Definitely! I was looking at the dispatching framework and wondering whether I get it right: the input array type determines which registered function is executed, right? So a numpy array would run with the scipy affine_transform, a cupy array with the cupy one. However, would the framework also allow running numpy arrays using cupy (after conversion / transfer to the GPU)? Or would the user in this case need to apply something like `image.map_blocks(lambda x: cupy.array(x))` to the array first?

jakirkham:
Yep that's right.

Dask tracks the array type contained within using something called the `meta` array. This is simply a small array used for testing functions on to see what they might produce and thus capture type changes. As a result Dask can pick up on this transformation of a NumPy array to a CuPy array. Here's an example to show this in action. Note `chunktype` is first a NumPy and then a CuPy array. No actual computing has happened Dask has just noticed the `meta` array would change and shared that info.

```python
In [1]: import numpy as np
   ...: import cupy as cp
   ...: 
   ...: import dask.array as da

In [2]: a = da.ones((10, 11), chunks=5)

In [3]: a
Out[3]: dask.array<ones, shape=(10, 11), dtype=float64, chunksize=(5, 5), chunktype=numpy.ndarray>

In [4]: a.map_blocks(cp.asarray)
Out[4]: dask.array<asarray, shape=(10, 11), dtype=float64, chunksize=(5, 5), chunktype=cupy.ndarray>
```

GenevieveBuckley:
> @jakirkham Definitely! I was looking at the dispatching framework and wondering whether I get it right: the input array type determines which registered function is executed, right? So a numpy array would run with the scipy affine_transform, a cupy array with the cupy one. However, would the framework also allow running numpy arrays using cupy (after conversion / transfer to the GPU)? Or would the user in this case need to apply something like `image.map_blocks(lambda x: cupy.array(x))` to the array first? @GenevieveBuckley

So I think what probably needs to happen is for us to make a function that dispatches *array creation*. Rather than creating numpy arrays and then transferring them over to the GPU (eg: for cupy arrays), it's probably a lot more efficient to just do the array creation once for the array type that matches the input data.

I was playing around with that idea here, if you want to take a look: https://github.com/dask/dask-image/pull/151/commits/acce4791a4b977358da0d0b20840501cfd4ee561

I didn't actually end up needing to use it for that PR, so that's why I took it out again. But we'll probably need something like this eventually, for exactly the problem you're having now.

m-albert:
> So I think what probably needs to happen is for us to make a function that dispatches array creation. Rather than creating numpy arrays and then transferring them over to the GPU (eg: for cupy arrays), it's probably a lot more efficient to just do the array creation once for the array type that matches the input data.

Hmm probably we should avoid initializing a full cupy array for the output with sth like `da.from_array(cp.zeros(output_shape), chunks=(100,))` as this would try to allocate memory for the full array on the GPU (at least on my machine it does). Instead, the currently implemented `da.zeros` could simply be the placeholder for both backends, as creation this way is very efficient. Alternatively, sth like this could be used, which probably isn't needed though:

```python
full_output_chunks = normalize_chunks(output_chunks, output_shape)
numblocks = tuple(map(len, full_output_chunks))
# initialize output array with reduced shape
output = da.zeros(numblocks, chunks=[1] * len(output_shape))
# regain full shape when performing transformation
output.map_blocks(..., chunks=full_output_chunks, dtype=output_dtype)
```

In general terms, I'm understanding that it would be useful to support these four operation modes:

*input image --processing backend--> output image*
1) np.array --scipy--> np.array
2) np.array --cupy---> np.array
3) np.array --cupy---> cp.array
4) cp.array --cupy---> cp.array

Actually, does `dask-image` have a way to support modes 2 and 3 for e.g. filter operations? Actually, do these even represent useful use cases? If not, the implemented dispatching mechanism `dask-image` could be enough to distinguish between cases 1) and 4) (if I understand it correctly).

Otherwise, if use cases 2 and 3 should be supported, we could introduce two additional arguments along these lines:
1) `processing_backend`: ['scipy', 'cupy'] (default 'scipy')
Based on this value, if required the input array would be converted into a dask array of type `cupy.ndarray` with `image.map_blocks(cp.asarray)`.
2) `output_type` [None, 'numpy', 'cupy'] (default None: take output from processing backend)
If required, the output would be converted back to chunktype `numpy.ndarray`.

The processing step could then be implemented using dispatching.

Let me know what you think!

m-albert:
Actually, defining all these different input / output cases probably doesn't help when the user can easily do that themselves using `image.map_blocks(asarray...)` and `output.map_blocks(cp.asnumpy)`.

So I added `cupy` capability by defining an `affine_transform` dispatcher and handled the array creation using a second `asarray` dispatcher in combination with explicitly passing a meta array to `map_blocks` that matches the input type.

It seems `numpy` and `cupy` tests are passing!

Here's a performance summary for
- input / output shape (768, 768, 768)
- input / output chunking (256, 256, 256)
- (16x Intel Xeon 3.7GHz and NVIDIA Titan RTX)
- `cupy` tests include copying data to and from GPU
- values in seconds

order | `scipy` | `dask-image` (numpy) | `dask-image` (cupy)
-- | -- | -- | --
0 | 23.1 | 3.7 | 2.1
1 | 32.8 | 5.2 | 4.5
3 | 123.5 | 15.6 | not supported by `cupy`

grlee77:
@m-albert: Can you specify which CuPy version you used for the benchmark above? (I ask because the `cupyx.scipy.ndimage` interpolation functions should be faster and more memory efficient in 8.0 than in prior releases)

m-albert:
@grlee77 for the benchmark above I used `cupy=6.0.0`. With `cupy=8.0.0` I get 2.7 and 2.0 seconds for orders 0 and 1 respectively.

m-albert:
@grlee77 Good point! I was using old `cupy=6.0.0` and updated the values now for `cupy=8.0.0`. With `order=1` the speed doubles, and the memory I didn't properly look at yet.

What I see though is that with consecutive runs the CUDA memory usage (as measured by `nvidia-smi`) steadily increases and doesn't seem to get released automatically.

GenevieveBuckley:
> Hmm probably we should avoid initializing a full cupy array for the output with sth like `da.from_array(cp.zeros(output_shape), chunks=(100,))` as this would try to allocate memory for the full array on the GPU (at least on my machine it does).

You're right, we should definitely not do this. But you can directly construct a dask array backed by numpy/cupy chunks using something along the lines of `da.zeros_like(image, shape=output_shape, dtype=image.dtype, chunks=output_chunks)`. This new feature in dask is very, very recent (https://github.com/dask/dask/pull/6680 was merged 6 days ago now) so you'll need the dask master branch if you want to try it out.

> In general terms, I'm understanding that it would be useful to support these four operation modes:
>
> *input image --processing backend--> output image*
>
> 1. np.array --scipy--> np.array
> 2. np.array --cupy---> np.array
> 3. np.array --cupy---> cp.array
> 4. cp.array --cupy---> cp.array

This isn't quite correct. What we do with the rest of the library is support two operation modes:

*input image --> output image*

1. dask array backed by numpy chunks --> dask array backed by numpy chunks
2. dask array backed by cupy chunks --> dask array backed by cupy chunks

There's no need or expectation that functions should do any type of converting. The only think you need to worry about as a developer here is that these two use cases are supported. I hope that helps simplify things a bit for you!

jni:
@GenevieveBuckley 

> So I think what probably needs to happen is for us to make a function that dispatches array creation. 

This might be of interest. Well, I'm pretty sure it is of interest since this exact same discussion is part of the motivation section:

https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html

m-albert:
> This isn't quite correct. What we do with the rest of the library is support two operation modes:
>
> input image --> output image
>
>dask array backed by numpy chunks --> dask array backed by numpy chunks
dask array backed by cupy chunks --> dask array backed by cupy chunks
There's no need or expectation that functions should do any type of converting. The only think you need to worry about as a developer here is that these two use cases are supported. I hope that helps simplify things a bit for you!

Yes you're right, thanks. That's the way it's implemented in the last commit.

> You're right, we should definitely not do this. But you can directly construct a dask array backed by numpy/cupy chunks using something along the lines of da.zeros_like(image, shape=output_shape, dtype=image.dtype, chunks=output_chunks). This new feature in dask is very, very recent (dask/dask#6680 was merged 6 days ago now) so you'll need the dask master branch if you want to try it out.

@GenevieveBuckley Very cool `dask` PR! `zeros_like` using dask's `broadcast_trick` makes it straight forward to instantiate `cupy` arrays that don't fit on the GPU.

In any case, for this PR, also to prevent depending on the latest `dask`, it'll be enough to inform `map_blocks` about the chunktype of the output array by manually passing a corresponding `meta` array as an argument. See the current implementation:

https://github.com/m-albert/dask-image/blob/4b108f148c1d7d303ec9de2f1e562fbecd6bdd87/dask_image/ndinterp/__init__.py#L120-L138

jakirkham:
> @GenevieveBuckley
> 
> > So I think what probably needs to happen is for us to make a function that dispatches array creation.
> 
> This might be of interest. Well, I'm pretty sure it is of interest since this exact same discussion is part of the motivation section:
> 
> https://numpy.org/neps/nep-0035-array-creation-dispatch-with-array-function.html

That's what Peter did recently in PR ( https://github.com/numpy/numpy/pull/16935 ). We are now working on using this in Dask ( https://github.com/dask/dask/pull/6738 ).

m-albert:
So I switched from using `map_blocks` to manually constructing the output dask array.

**Problem**
I had realised that computations involving input arrays larger than memory would lead to a `MemoryError`. This was the case because the entire input array was computed and passed to each `resample_chunk` task. Meaning that so far the function had blocked one of the main advantages of using dask arrays.

**Solution**
Of course typically, `map_blocks` directly provides the necessary computed blocks to the inner function, however in this case the inner function first needs to find out which input region to work on. Therefore the input/output mapping needs to be determined prior to the actual computation. This is now performed at the top level in `ndinterp.affine_transform`, followed by a manual construction of the dask graph computing the output array (using `dask.highlevelgraph.HighLevelGraph`).

**Result**
The result is that `ndinterp.affine_transform` function doesn't need to compute the entire input array anymore allowing it to deal with large input arrays. I added tests that make sure this is given (by simply checking that no error is raised in the case of computing a small output from huge inputs, maybe there's a better way).

@jakirkham @GenevieveBuckley let me know if you can think of further improvements!

dask_image/dispatch/_dispatch_ndinterp.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @dispatch_affine_transform.register(cupy.ndarray)
+    def cupy_affine_transform(*args, **kwargs):
+
+        return cupyx.scipy.ndimage.affine_transform
+
+
+dispatch_asarray = Dispatcher(name="dispatch_asarray")
+
+# ================== affine_transform ==================
```
GenevieveBuckley, line 35:
```suggestion
# ================== asarray ==================
```

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+    # Further parameter checks are performed directly by
+    # `ndimage.affine_transform`.
+
+    matrix = np.asarray(matrix)
+    offset = np.asarray(offset).squeeze()
+
+    # these lines were copied and adapted from `ndimage.affine_transform`
+    if (matrix.ndim == 2 and matrix.shape[1] == image.ndim + 1 and
+            (matrix.shape[0] in [image.ndim, image.ndim + 1])):
+        # if matrix.shape[0] == image.ndim + 1:
```
GenevieveBuckley, line 100: I notice lines 100-107 are commented out, do these need to be removed?

dask_image/ndinterp/__init__.py
```diff
@@ -0,0 +1,236 @@
...
+            rel_image_i[dim] = np.clip(rel_image_i[dim], 0, s - 1)
+            rel_image_f[dim] = np.clip(rel_image_f[dim], 0, s - 1)
+
+        rel_image_slice = tuple([slice(int(rel_image_i[dim]),
+                                       int(rel_image_f[dim]) + 1)
+                                 for dim in range(n)])
+
+        rel_image = image[rel_image_slice]
+
+        # Modify offset to point into cropped image.
```
GenevieveBuckley, line 196: Here lines 196-204 are also commented out. Do these need to be removed?

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,314 @@
...
+import dask_image.ndinterp as da_ndinterp
+
+import numpy as np
+import dask.array as da
+from scipy import ndimage
+
+
+class Helpers:
+    @staticmethod
+    def test_affine_transform(n=2,
```
GenevieveBuckley, line 15: A class with a single static method probably would be simpler if it was a single function. 

If you call that function something like `validate_affine_transform` (something without the `test_` prefix so pytest doesn't pick it up) then you can get rid of both the `Helper` class and the pytest fixture in one go.

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -0,0 +1,314 @@
...
+    # fully computed, this array would occupy 8TB
+    image = da.random.random([10000] * 3, chunks=(200, 200, 200))
+    image_t = da_ndinterp.affine_transform(image, np.eye(3), [0, 0, 0],
+                                           output_chunks=[1, 1, 1],
+                                           output_shape=[1, 1, 1])
+
+    # if more than the needed chunks should be computed,
+    # this would take long and eventually raise a MemoryError
+    # use pytest-timeout?
+    image_t[0, 0, 0].compute()
```
GenevieveBuckley, line 243: Yes, pytest-timeout sounds like a good idea here

dask_image/ndinterp/__init__.py
m-albert: Yes thanks, I had accidentally left them there. That warning is instead raised by the `ndimage` test run in line 84:

```python
    ndimage_affine_transform(np.zeros([0] * image.ndim),
                             matrix,
                             offset)
```

dask_image/ndinterp/__init__.py
m-albert: Those intend to explain how `offset` is transformed into `offset_prime` for each output chunk.

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
m-albert: Oh I wasn't aware that changing the prefix was enough 😅. Thanks!

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
m-albert: Great. The test `test_affine_transform_large_input_small_output_cpu` takes around 2-3s on my macbook, so I set the timeout at 15s for now to be on the safe side.

And added a `pytest-timeout` requirement in `setup.py`.

```
test_requirements = [
    "flake8 >=3.4.1",
    "pytest >=3.0.5",
    "pytest-flake8 >=0.8.1",
    "pytest-timeout >=1.0.0",
    "tifffile >=2018.10.18",
]
```

GenevieveBuckley:
I think we are very close to merging this!

One of the last remaining things I would like to do is to play with the new feature. Did you have a specific example you were using when you developed this? (Bonus: we might also be able to add the example to the documentation later on as well)

m-albert:
@GenevieveBuckley I needed chunkwise affine transforms for fusing multi-view light sheet image data (see [here](https://github.com/m-albert/MVRegFus)).

Now that `napari` supports affine transforms I wanted to use this for visualizing multi-view registration and fusion, so maybe I could eventually write up an example in this context :)

GenevieveBuckley:
> 
> 
> @GenevieveBuckley I needed chunkwise affine transforms for fusing multi-view light sheet image data (see [here](https://github.com/m-albert/MVRegFus)).
> 
> Now that `napari` supports affine transforms I wanted to use this for visualizing multi-view registration and fusion, so maybe I could eventually write up an example in this context :)

That would be great!

sofroniewn:
This is amazing @m-albert, very excited about this!

To compare this with napari we should probably be able to do something like
```python
import napari
from dask_image.ndimage import affine_transform

with napari.gui_qt():
    viewer = napari.Viewer()
    viewer.add_image(data, affine=matrix, colormap='green')
    viewer.add_image(affine_transform(data, matrix), colormap='magenta')
```
where `data` is a dask array.

m-albert:
@sofroniewn So great that `napari` supports affine transforms now!

Here's the direct comparison of the affine transforms as computed by `scipy` and `dask-image` to `napari`s using the code you suggested:

```python
import numpy as np
import napari
from dask_image.ndinterp import affine_transform as affine_transform_dask
from scipy.ndimage import zoom, affine_transform as affine_transform_scipy
import dask.array as da

data = np.random.random((7, 7, 7))
data = zoom(data, 20)
data = da.from_array(data, chunks=(20, 20, 20))

from scipy.spatial.transform import Rotation as R
r = R.from_rotvec(np.pi/3. * np.array([1, 0, 0]))

im_center = np.array(data.shape) / 2.

matrix = np.eye(4)
matrix[:3, :3] = r.as_matrix()
matrix[:3, 3] = -np.dot(r.as_matrix(), im_center) + im_center

with napari.gui_qt():
    viewer = napari.Viewer()
    viewer.add_image(data, colormap='gray', opacity=0.3, name='orig')
    viewer.add_image(data, affine=matrix, colormap='red', name='napari-affine')
    viewer.add_image(data, affine=np.linalg.inv(matrix), colormap='green', name='napari-affine-inv')
    viewer.add_image(affine_transform_dask(data, matrix, offset=[0, 0, 0]), colormap='blue', name='ndimage-affine')
    viewer.add_image(affine_transform_scipy(data, matrix, offset=[0, 0, 0]), colormap='magenta', name='dask-image-affine')
```

Turns out that the conventions are inverted in the sense that while in the case of `scipy.ndimage.affine_transform` the output is associated to the domain of the linear transform y=Mx+c (and the image input to the codomain), in the case of `napari` it's the opposite. Therefore inverting the homogeneous transform matrix using `np.linalg.inv` is required for obtaining overlapping output images.

sofroniewn:
Awesome, good to know about the inverse difference. We can think about that more on the napari side if that's something we might want to revisit at some point!!

GenevieveBuckley:
Ok, it's time to test this feature out in the wild 😄 

Fantastic to have your contribution here @m-albert, thank you!

m-albert:
Great, thanks @GenevieveBuckley and everyone here for all the ideas and feedback!

martinschorb:
Hi,

I wonder if the higher-level functions in `ndimage` that implicitely use `affine_transform` (like rotate) can be made to work with your approach Marvin as well.

Taht would be really awesome.

jakirkham:
@martinschorb, would suggest raising a new issue to discuss that 🙂

Issue #161, opened 2020-10-20
imread - investigate possible performance improvement


GenevieveBuckley:
It's been found that the performance of `da.map_blocks` is much better than `da.stack` when joining large arrays: https://github.com/dask/dask/issues/5913

It's unclear if `da.concatenate` ([like we use in imread](https://github.com/dask/dask-image/blob/master/dask_image/imread/__init__.py#L89)) is also slower, but this seems likely. We should investigate if we can get a performance benefit by switching to `da.map_blocks`.

GenevieveBuckley:
https://github.com/dask/dask/issues/5913 has several comments describing how to accurately profile speedups

GenevieveBuckley:
One potential blocker: jni thinks the reason `da.stack` is slow is because it allows the flexibility of unequal chunk sizes. I can't remember, but I don't think we support folders full of *differently sized* images? If I'm wrong about that it might mean we don't want to give up that functionality for some speed improvement in other cases.

@jni says [here](https://github.com/napari/tutorials/issues/93#issuecomment-712669205):

> For reference, a couple of usage examples of `map_blocks`:
> 
> https://github.com/jni/napariboard-proto/blob/4b00303e567328270e02bd7e977cd2656040f8ea/napariboard.py#L70-L107
> 
> and
> 
> https://github.com/jni/napari-demos/blob/91b5e8ab03852fc93fe6d74d81fed2d39dbdda3b/rootomics.py#L20-L54
> 
> @GenevieveBuckley it would be good to know whether `da.concatenate`, used in `dask_image.imread`, suffers from the same performance issues as `da.stack`. My suspicion is that it does, because if I understand correctly, the performance issue comes from `da.stack` allowing unequal chunk sizes along an axis.

jakirkham:
One thing worth noting (and I may have mentioned this to Juan as well) is `map_blocks` will fill in certain special keywords of the user-defined function if the function takes them. One of these is [`block_info` (need to scroll down a bit)]( https://docs.dask.org/en/latest/array-api.html?highlight=block_info#dask.array.map_blocks ). This would help provide positioning info like where in the larger Dask Array this is, how big this chunk should be, etc. May be helpful when figuring out what data should be filled in for a specific chunk.

m-albert:
Hi guys, just saw this here and remembered that I did stumble upon bad performance of `dask.image.array.stack` in the past when creating large arrays.

> We should investigate if we can get a performance benefit by switching to da.map_blocks.

[...]

Issue #160, opened 2020-10-12
Failing tiff imread tests


GenevieveBuckley:
It looks like our imread test are failing, as shown by the CI for this unrelated PR https://github.com/dask/dask-image/pull/159

Here's a link to the output of one of the CI jobs showing this failure:
https://app.circleci.com/pipelines/github/dask/dask-image/49/workflows/9b4326b7-fc8a-433f-ac0c-ce36cc344e7f/jobs/1315

jakirkham:
Based on the failure, it seems `nframes` isn't being respected when chunking the data.

PR #162, opened 2020-10-21
Fix imread tests and cover warnings

m-albert:
This PR intends to fix currently failing imread tests https://github.com/dask/dask-image/issues/160.

Locally, I found that the problem with the current test `test_imread.test_core.py::test_tiff_imread` was that in case of `shape[0] > 1`, only one image file was written and read. I fixed this by changing

```python
    fn = str(dirpth.join("test.tiff"))
    with tifffile.TiffWriter(fn) as fh:
        for i in range(len(a)):
            fh.save(a[i])
```

into
```python
    fn_pattern = str(dirpth.join("test_%s.tiff"))
    for i in range(shape[0]):
        with tifffile.TiffWriter(fn_pattern % i) as fh:
            fh.save(a[i])

    if shape[0] == 1:
        fn = fn_pattern % 0
    else:
        fn = fn_pattern % '*'
```
(edited)

Splitting up the cases `shape[0]==1` and `>1` seemed necessary because `pims.open` doesn't seem to find single files using the wildcard `*`.

This fixes my tests locally, though I'm not quite sure yet why the CI currently passes for py36 (as seen e.g. here https://github.com/dask/dask-image/pull/159).

Aiming to have this merged before submitting the PR addressing the proposed `imread` change in https://github.com/dask/dask-image/issues/161.

### DIFF
--- a/tests/test_dask_image/test_imread/test_core.py
+++ b/tests/test_dask_image/test_imread/test_core.py
@@ -36,17 +36,17 @@ def test_errs_imread(err_type, nframes):
     ]
 )
 @pytest.mark.parametrize(
-    "nframes, shape",
+    "nframes, shape, runtime_warning",
     [
-        (1, (1, 4, 3)),
-        (-1, (1, 4, 3)),
-        (3, (1, 4, 3)),
-        (1, (5, 4, 3)),
-        (2, (5, 4, 3)),
-        (1, (10, 5, 4, 3)),
-        (5, (10, 5, 4, 3)),
-        (10, (10, 5, 4, 3)),
-        (-1, (10, 5, 4, 3)),
+        (1, (1, 4, 3), None),
+        (-1, (1, 4, 3), None),
+        (3, (1, 4, 3), "`nframes` larger than"),
+        (1, (5, 4, 3), None),
+        (2, (5, 4, 3), "`nframes` does not nicely divide"),
+        (1, (10, 5, 4, 3), None),
+        (5, (10, 5, 4, 3), None),
+        (10, (10, 5, 4, 3), None),
+        (-1, (10, 5, 4, 3), None),
     ]
 )
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Nice detective work!

It seems there was a change introduced between `tifffile` version 2020.8.25 and 2020.9.3 that led to this behaviour. We pin almost everything in our test environments, but we don't seem to have pinned `tifffile`.

tests/test_dask_image/test_imread/test_core.py
```diff
@@ -69,12 +69,19 @@ def test_tiff_imread(tmpdir, seed, nframes, shape, dtype):
 
     a = np.random.uniform(low=low, high=high, size=shape).astype(dtype)
 
-    fn = str(dirpth.join("test.tiff"))
-    with tifffile.TiffWriter(fn) as fh:
-        for i in range(len(a)):
+    fn_pattern = str(dirpth.join("test_%s.tiff"))
+    for i in range(shape[0]):
+        with tifffile.TiffWriter(fn_pattern % i) as fh:
```
GenevieveBuckley, line 74: This looks as if the previous tests were writing a multi-page tiff to a single file, and the new test is writing many individual tiff files instead. Do I misunderstand this?

If so, that seems like a significant change to me, and one that reduces the amount of meaningful test coverage we have.

GenevieveBuckley: Update: I've just tried this on my local machine and yes, I think there is a meaningful difference between these, they're not equivalent.

GenevieveBuckley:
I suspect if we add `contiguous=True` to our tifffile save command in the tests, that might fix things for us.

GenevieveBuckley:
> 
> 
> I suspect if we add `contiguous=True` to our tifffile save command in the tests, that might fix things for us.

It works, hooray! If all the CI tests go well, I'll merge https://github.com/dask/dask-image/pull/164 with that fix soon.

I really like the way you have checked that the appropriate runtime warnings were raised in this PR, so I've included the same thing there too.

GenevieveBuckley:
We've merged a fix! 

I'm going to close this now, thanks for doing this work 😄

m-albert:
> I suspect if we add contiguous=True to our tifffile save command in the tests, that might fix things for us.

Great that you found the root cause of the problems :)

Thanks for fixing this!

PR #168, opened 2020-10-29
Update dependencies in RTD environment

jakirkham:
ReadTheDocs builds recently started failing. This seems to be due to a `pip` option being passed that the version of `pip` we use here doesn't understand. In an attempt to fix this issue, try modernizing our dependencies here.

https://readthedocs.org/projects/dask-image/builds/12159489/

### DIFF
--- a/environment_doc.yml
+++ b/environment_doc.yml
@@ -4,12 +4,12 @@ channels:
   - conda-forge
 
 dependencies:
-  - pip==18.0
-  - wheel==0.31.1
-  - sphinx==1.7.5
-  - dask-sphinx-theme==1.1.0
-  - dask==0.16.1
-  - numpy==1.11.3
+  - pip==19.0.1
+  - wheel==0.32.3
+  - sphinx==3.2.1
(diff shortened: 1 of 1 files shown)


jakirkham:
Hmm...the cause of the issue appears to be ( https://github.com/readthedocs/readthedocs.org/issues/7498 ). This won't fix it, but may be a good idea anyways.

jakirkham:
Ok the latest commit should ensure we have a version of `pip` that has the flag RTD is using. Namely RTD is trying to use the [new `pip` resolver]( https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-2-2020 ). That said, I'm not totally sure we want `pip` to be changing the environment after Conda has constructed it in a particular way.

GenevieveBuckley:
Thanks for looking into this John!

jakirkham:
Let's give this a try 🍀

jakirkham:
Looks like it [passed]( https://readthedocs.org/projects/dask-image/builds/12222842/ )! 🎉

Issue #163, opened 2020-10-22
Join the big image sprint at SciPy Japan 2020


GenevieveBuckley:
dask-image will join a "big image" sprint at SciPy Japan next week.  There's more detail about the sprint topics here: https://hackmd.io/qpiRFdIsTdCAfdr_C2mMUA

The sprint will be held remotely on November 1-2, 9am-5pm Tokyo time. Anyone who wants to participate is welcome to join. We welcome asynchronous participation too, if the time difference is not a friendly one for you.

Let me know if you want to be involved!

m-albert:
@GenevieveBuckley I'd be super happy to participate :)

Unfortunately the schedule translates to 1am-9am in Zurich. I could join at the beginning and / or towards the end. Or asynchronously!

jni:
@m-albert async works! 😊 Also @snowformatics might be joining from Europe, probably working on jni.github.io/skan, in case you have some interest in skeletons. =) See jni/skan#96

GenevieveBuckley:
Asynchronous is great! It'd be great to have you around @m-albert 😄

GenevieveBuckley:
Here's a copy of the list of possible sprint ideas from the hackmd page:
- [ ] FEATURE: add deconvolution implementation
     * https://github.com/dask/dask-blog/pull/77
     * https://github.com/dask/dask-image/issues/149
- [ ] FEATURE: pathlib input for imread functions
    * https://github.com/dask/dask-image/issues/139
- [ ] MAINTENANCE: rename `sum` to `sum_labels` and add a deprecation warning
    * https://github.com/dask/dask-image/issues/123
- [ ] TUTORIALS: Turn example pipeline from Genevieve's 2020 talks into a PR for the dask-examples repository.
    * https://github.com/GenevieveBuckley/dask-image-talk-2020
    * https://github.com/dask/dask-examples
- [ ] TUTORIALS: Clarify dask best practices in napari tutorials.
    - [ ] Update [applications/dask.md](https://github.com/napari/tutorials/blob/master/applications/dask.md) to add a similar note like this https://github.com/dask/dask-blog/pull/76
    - [ ] Simplify dask in example #1: https://github.com/napari/napari/pull/1610
    - [ ] Simplify dask in example #2: https://github.com/napari/tutorials/pull/64
[...]

Issue #147, opened 2020-06-08
Running test suite as part of PyPI release


jakirkham:
As part of the PyPI release process, I'm wondering if we should run the test suite. Maybe this would help us catch some of the issues before publishing.

PR #172, opened 2020-11-01
MAINTENANCE: rename sum to sum_labels and add a deprecation warning #123

tkoyama010:
MAINTENANCE: rename sum to sum_labels and add a deprecation warning #123

Join the big image sprint at SciPy Japan 2020 (#163)

PR #175, opened 2020-11-02
Running test suite as part of PyPI release #147

tkoyama010:
Running test suite as part of PyPI release #147
This is a sprint contribution in ScipyJapan2020 (#163) .

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -6,6 +6,9 @@ name: Upload Python Package
 on:
   release:
     types: [created]
+  pull_request:
+    branches:
+      - master
 
 jobs:
   deploy:
@@ -22,7 +25,11 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
+    - name: Test
+      run: |
+        make test
     - name: Build and publish
+      if: contains(contains(github.ref, 'master')
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -24,7 +24,7 @@ jobs:
         pip install setuptools wheel twine
     - name: Test
       run: |
-        make test
+        python setup.py test
     - name: Build and publish
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -6,9 +6,6 @@ name: Upload Python Package
 on:
   release:
     types: [created]
-  pull_request:
-    branches:
-      - master
 
 jobs:
   deploy:
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -26,7 +26,6 @@ jobs:
       run: |
         make test
     - name: Build and publish
-      if: contains(github.ref, 'master')
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -29,7 +29,7 @@ jobs:
       run: |
         make test
     - name: Build and publish
-      if: contains(contains(github.ref, 'master')
+      if: contains(github.ref, 'master')
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/python-publish.yml
```diff
@@ -22,7 +25,11 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
+    - name: Test
+      run: |
+        make test
     - name: Build and publish
+      if: contains(contains(github.ref, 'master')
```
tkoyama010, line 32:
```suggestion
      if: contains(github.ref, 'master')
```

tkoyama010:
Oh my. Test script which we created in #172 seems not working :cry:
https://github.com/dask/dask-image/pull/175/checks?check_run_id=1340168042

tkoyama010:
Before reviewing this Pull Request, please review see #176 :bow:

.github/workflows/python-publish.yml
```diff
@@ -6,6 +6,9 @@ name: Upload Python Package
 on:
   release:
     types: [created]
+  pull_request:
+    branches:
+      - master
```
tkoyama010, line 11:
```suggestion
```

.github/workflows/python-publish.yml
```diff
@@ -22,7 +25,11 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
+    - name: Test
+      run: |
+        make test
     - name: Build and publish
+      if: contains(github.ref, 'master')
```
tkoyama010, line 32:
```suggestion
```

tkoyama010 (commented):
:wastebasket: Remove patch of check

tkoyama010:
Now it is ready for review :rocket:

.github/workflows/python-publish.yml
```diff
@@ -22,7 +25,11 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
+    - name: Test
+      run: |
+        make test
```
GenevieveBuckley, line 27: `make test` seems equivalent to `python setup.py test`, which is now deprecated(https://github.com/pypa/setuptools/issues/1684). 

I get a warning when I run this, see: https://github.com/GenevieveBuckley/dask-image/runs/1340523683?check_suite_focus=true

```
Run make test
python setup.py test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

If possible, I'd like us to replace this with something that is less likely to stop working at some future point in time.

.github/workflows/python-publish.yml
GenevieveBuckley: I think we can remove this line, the release tag is already attached to the master branch.

.github/workflows/python-publish.yml
```diff
@@ -6,6 +6,9 @@ name: Upload Python Package
 on:
   release:
     types: [created]
+  pull_request:
```
GenevieveBuckley: This should be:
```on:
  release:
    types: [created]
```

Not this:
```
on:
  release:
    types: [created]
  pull_request:
    branches:
      - master
```

GenevieveBuckley: I understand you may want to change this to test the github action is working on your branch before we merge this, of course!

GenevieveBuckley (changes requested):
Thanks for making this PR @tkoyama010 

I think we should find an alternative to `make test` / `python setup.py test` since that command is now deprecated (https://github.com/pypa/setuptools/issues/1684). Other than that, I like this a lot 😄

.github/workflows/python-publish.yml
tkoyama010: Yes I will remove this when we will merge this PR. Thanks for comment.

.github/workflows/python-publish.yml
tkoyama010: You are right. Thanks for comment.

.github/workflows/python-publish.yml
tkoyama010: Thanks. I am +1 for this.

tkoyama010:
```suggestion
        python setup.py test
```

tkoyama010:
Thanks @GenevieveBuckley for your review.

GenevieveBuckley:
I'm going to close this PR in favour of https://github.com/dask/dask-image/pull/291 
The advantage with that is:
* there will be no difference between the regular test CI, and the tests that are run just before the new release is uploaded to PyPI
* it will also test all operating systems (Windows, Mac, and Linux) instead of just on Linux

GPUtester:

GenevieveBuckley:
Thanks again for all your help @tkoyama010!

Issue #181, opened 2021-01-19
dask_image imread performance issue


jmontoyam:
Dear dask_image community,

I am a new dask_image user. Maybe, due to my beginner level, I am doing something wrong, but I noticed that reading a collection of images using dask_image is much slower that using single-threaded skimage. I have installed the latest dask_image version available in pypi (dask_image version 0.4.0). In the following example, I am reading 398 images, all of them with the same dimension (64x10240, uint16). Taking into account the dimensions and numbers of images, I would expect dask_image to be slighly slower that single-threaded skimage (due to the tiny dask overhead involved in opening this small number of "tiny images"), but instead the performance of dask_image is much slower (around 24x). Then I proceed to implement the image reading function in pure-dask and the performance is much better than the one obtained with dask_image. In the following I will report the benchmarks results (all the following code-snippets load the same data successfully):

import glob
import numpy as np
import skimage.io
import dask_image.imread
from dask import delayed
import dask.array as da

---------------------------------------------------------------------------------------
### Single-threaded-skimage baseline
```python
%%time
all_images = sorted(glob.glob(f"{path_images}/*.tif"))
array_images = np.zeros((len(all_images), 64, 10240), dtype=np.uint16)
for idx, image in enumerate(all_images):
    array_images[idx] = skimage.io.imread(image)
```
Elapsed time: 510 milliseconds

---------------------------------------------------------------------------------------
### Using dask_image
```python
%%time
using_dask_image = dask_image.imread.imread(f"{path_images}/*.tif")
array_dask_image = using_dask_image.compute()
```
Elapsed time: 12.1 seconds

---------------------------------------------------------------------------------------
### Using pure-dask
```python
%%time
lazy_imread = delayed(skimage.io.imread)  # lazy reader
lazy_arrays = [lazy_imread(image) for image in all_images]
dask_arrays = [
    da.from_delayed(delayed_reader, shape=(64, 10240), dtype=np.uint16)
    for delayed_reader in lazy_arrays
]
using_dask = da.stack(dask_arrays, axis=0).compute()
```
Elapsed time: 1.09 seconds

---------------------------------------------------------------------------------------
### Using dask-image with synchronous scheduler
```python
%%time
array_dask_image = using_dask_image.compute(scheduler="synchronous")
```
Elapsed time: 3 seconds

---------------------------------------------------------------------------------------
### Using dask-image with processes scheduler
```python
%%time
array_dask_image = using_dask_image.compute(scheduler="processes")
```
[...]

Issue #121, opened 2019-08-01
Loading stack of images slow


jlevy44:
I have about 20k 500px500p images stored in a directory.

My ultimate goal is to reshape them into only a few images, blocking the 500s into the few larger images.

I first wanted to see how fast I could load the images.
```
arr=dask_image.imread.imread('*.png')[:100].compute()
```
versus
```
from PIL import Image
arr=[Image.open(img) for img in glob.glob('*.png')[:100]]
```
The second task is performing much much much faster than the first task (first task hangs). 

Can you help me understand why this is so and what I can do to speed this up?

Chunk size is (1,500,500,3) but I changed it to (25,500,500,3) and it didn't really make a difference. I would normally expect this to be a fast and easy Dask operation. I even started processing and threaded distributed clients with no luck. 

Help is very much appreciated. Thanks!

jlevy44:
Ideally, I'd like to be able to load the 20k photos and reshape in minutes, not hours.

Then conversion of the larger arrays to zarr for permanent storage with lower filesystem impact.

jlevy44:
I've found this to be pretty fast as well:
```
arr3=da.compute(*[dask.delayed(lambda x: np.array(Image.open(x)))(img) for img in imgs[:100]])
```
What am I doing wrong with my first command?

jlevy44:
Also, scikit-image's ImageCollection also performs the load really fast.

```
ImageCollection(imgs2[:100])
```

jlevy44:
And so far the fastest performing:
```
arr=da.from_array([da.compute(*[dask.delayed(lambda x: np.array(Image.open(x)))(img) for img in imgs[i:i+1000]]) for i in list(range(10000))[::1000]])
```

jlevy44:
It does autochunk to:
```
dask.array<array, shape=(10, 1000, 500, 500, 3), dtype=uint8, chunksize=(10, 125, 125, 125, 3)>
```
But perhaps this partitioning is advantageous?

jni:
@jlevy44 sorry about the slow response here. Any reason why you closed this? It certainly might need followup!

jlevy44:
Oh, we can reopen. Although I had solved the problem for myself on my end, I am certainly open to discussion to benefit the greater community.

jni:
Nice, thanks! =) Was the solution you went with the one you noted as the fastest above?

jakirkham:
[...]

PR #182, opened 2021-01-20
Imread performance: reduced overhead of pim.open calls when reading from image sequence

m-albert:
As reported by https://github.com/dask/dask-image/issues/181 the performance of `dask_image.imread` is bad in the case of reading from many input files. Also related are https://github.com/dask/dask-image/issues/121 https://github.com/dask/dask-image/issues/161.

One problem with the current implementation that I noticed is that when calling `dask_image.imread.imread` with a file pattern such as `im_*.tif`, for each tile that is loaded `pims.open` is called on the entire file pattern. This then leads to many unnecessary instantiations of `pims.ImageSequenceND`s producing a large overhead.

In this proposed fix I use glob to match filenames and frames to call `pims.open` only on the files that are actually being loaded.

### DIFF
--- a/dask_image/imread/__init__.py
+++ b/dask_image/imread/__init__.py
@@ -5,7 +5,7 @@ __author__ = """John Kirkham"""
 __email__ = "kirkhamj@janelia.hhmi.org"
 
 
-import itertools
+import glob
 import numbers
 import warnings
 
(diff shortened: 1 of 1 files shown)


m-albert:
Here some performance tests (inspired by https://github.com/dask/dask-image/issues/181):

Data prep:
```python
import numpy as np
import skimage.io
import dask_image.imread
import glob
from dask import delayed
import dask.array as da

Nz, Ny, Nx = 1000, 100, 100
im = np.random.randint(0, 1000, (Nz, Ny, Nx))

for z in range(Nz):
    skimage.io.imsave('data/im_%05d.tif' %z, im[z])
```

Timings:
```python
%%time
all_images = sorted(glob.glob("data/im_*.tif"))
imgs = []
for idx, image in enumerate(all_images):
    imgs.append(skimage.io.imread(image))
imgs = np.array(imgs)
```
**Serial read without dask, 420ms**

```python
%%time
lazy_imread = delayed(skimage.io.imread)  # lazy reader
lazy_arrays = [lazy_imread(image) for image in all_images]
dask_arrays = [
    da.from_delayed(delayed_reader, shape=(Ny, Nx), dtype=np.uint16)
    for delayed_reader in lazy_arrays
]
using_dask = da.stack(dask_arrays, axis=0).compute()
```
**Using dask delayed: 1.05s**

```python
%%time
lazy_imread = delayed(skimage.io.imread)  # lazy reader
lazy_arrays = [lazy_imread(image) for image in all_images]
dask_arrays = [
    da.from_delayed(delayed_reader, shape=(Ny, Nx), dtype=np.uint16)
    for delayed_reader in lazy_arrays
]
using_dask = da.stack(dask_arrays, axis=0).compute()
```
**Master: 16s
This PR: 1.04s**

GenevieveBuckley:
Genuinely sorry, not sure how this managed to fall off my radar. Adding it back to the to-do list now!

GenevieveBuckley:
Seems like you already have one review from John - https://github.com/dask/dask-image/issues/181#issuecomment-765319349

> Thank you very much for the explanation @m-albert !...By the way, I have checked the #182 , and everything looks good to me,
> thank you very much for taking the time to address this issue! ;)

GenevieveBuckley:
Looks good to me too. 

At some point we should consider adding performance tests, but that's a conversation for another day.

m-albert:
@GenevieveBuckley Great, thanks for reviewing and merging!

Issue #204, opened 2021-04-20
`affine_transform` fails in simplest possible use case


martinschorb:
**What happened**:
Running `affine_transform` on a simple 2D image produces a `TypeError` when `offset` is *None*

```python
from scipy import  misc
from dask_image.ndinterp import affine_transform

img = da.from_array(misc.ascent())

matrix = np.array([[0,-1],[1,0]])

img_45 = affine_transform(img, matrix)
```
results in:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-28-84fbda5afee9> in <module>
      8 matrix = np.array([[0,-1],[1,0]])
      9 
---> 10 img_45 = affine_transform(img, matrix)
     11 #plt.imshow(img_45)

/g/emcf/software/python/miniconda/envs/xtomo/lib/python3.8/site-packages/dask_image/ndinterp/__init__.py in affine_transform(image, matrix, offset, output_shape, order, output_chunks, **kwargs)
    154             rel_image_edges = matrix * out_chunk_edges + offset
    155         else:
--> 156             rel_image_edges = np.dot(matrix, out_chunk_edges.T).T + offset
    157 
    158         rel_image_i = np.min(rel_image_edges, 0)

TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

**Anything else we need to know?**:

The problem is due to `offset` still being `None`. In this case it is never touched, and `offset = np.asarray(offset).squeeze()` in l. 96 also yields `None`.

**Environment**:

- Dask version: 2021.3.0 
- dask-image                0.5.0  
- Python version:    3.8.8
- Operating System:  CentOS 7
- Install method (conda, pip, source): conda

martinschorb:
Also, I found that `output_shape` needs to be a tuple in dask-image, whereas the "original" function also accepts `np.array`s.

Otherwise l.126:  `normalized_chunks = da.core.normalize_chunks(output_chunks, output_shape)` fails.

martinschorb:
In addition, the variable that is returned cannot be specified as `output` kwarg. 

I now have handcrafted my own `rotate` around it.

martinschorb:
> In addition, the variable that is returned cannot be specified as `output` kwarg.

This seems to be more tricky in cases with ndim>2 as this `output` accepts not only arrays but also slice iterators at arbitrary angles and positions pointing to an array. 
[...]

PR #205, opened 2021-04-20
affine_transform: Remove inconsistencies with ndimage implementation.

m-albert:
As reported by @martinschorb in https://github.com/dask/dask-image/issues/204, there are some inconsistencies in `dask_image.ndinterp.affine_transform` with respect to the `ndimage` implementation.

This PR
1) fixes the behaviour when not providing an offset argument and adds a test for this.
2) adapts the function API to expect tuples of ints for the shape related parameters, just as the downstream `scipy` implementation does. As @martinschorb points out, the `scipy` implementation also works when providing `output_shape` as a `ndarray`, so I now explicitly perform a conversion into tuple within the function to also allow for this (please object if this doesn't make sense).

I don't see reasons for why the inconsistencies had been there in the first place and assume I had simply not been careful enough during the implementation. Also, there should be no backwards compatibility problem introduced by this PR.

### DIFF
--- a/dask_image/ndinterp/__init__.py
+++ b/dask_image/ndinterp/__init__.py
@@ -25,7 +25,7 @@ __all__ = [
 def affine_transform(
         image,
         matrix,
-        offset=None,
+        offset=0.0,
         output_shape=None,
         order=1,
         output_chunks=None,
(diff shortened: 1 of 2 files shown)


m-albert:
As @martinschorb mentioned, another inconsistency with the `scipy` implementation is that currently `dask_image.ndinterp.affine_transform` does not support providing the output array as the scipy `output` argument (this is actually the case for most `dask_image` functions). Instead a new array is always returned.

If I understand things correctly, providing a dask array as an `output=` doesn't provide any benefits as the returned numpy chunks would always be new copies (preventing in-place operations). However @martinschorb mentioned that it could be useful to be able to modify only slices of an array. While I'm not sure if this is currently possible, the newly implemented item assignment for dask arrays could enable this soon: https://github.com/dask/dask/issues/7392 and https://github.com/dask/dask/issues/7392.

GenevieveBuckley (approved):

PR #215, opened 2021-05-06
Add spline_filter and spline_filter1d

grlee77:
This PR adds the spline prefiltering functions that are correspond to:
`scipy.ndimage.spline_filter`
`scipy.ndimage.spline_filter1d`

The periodic modes `"wrap"` and `"grid-wrap"` are not currently supported, but all other modes are.

The one empirical nature of this implementation is the equation I came up with for estimating the overlap (depth) needed based on the interpolation order. In practice these are infinite impulse response (IIR) filters, but in practice the error terms decay pretty quickly with distance from an edge and the rate of decay is related to the magnitude of the filter poles. The current settings are enough for tests to pass at the tolerance of 1e-6 used in the tests. With the current default, the `depth` currently ends up being 11, 14, 19 and 22 for spline `order` of 2, 3, 4 and 5, respectively. I exposed a `depth` keyword-only argument in case users want to override this setting, but am not certain if that is desirable. 

This PR also enables use of `prefilter=True` in the existing `affine_transform` function when `mode='constant'`. This is needed to get good results with order > 1. Although `reflect` and `mirror` are also supported by the `spline_filter` functions added here, they are not currently supported in `affine_transform` itself. I have not yet tried to support `mode='nearest'` or `mode='grid-constant'` for use within `affine_transform` because those two cases do not have an analytic solution implemented and involve the need for temporary internal padding in SciPy's own implementation, making it more difficult to adapt here.

### DIFF
--- a/dask_image/dispatch/_dispatch_ndinterp.py
+++ b/dask_image/dispatch/_dispatch_ndinterp.py
@@ -30,6 +30,44 @@ def register_cupy_affine_transform():
         return cupyx.scipy.ndimage.affine_transform
 
 
+dispatch_spline_filter = Dispatcher(name="dispatch_spline_filter")
+
+# ================== spline_filter ==================
+@dispatch_spline_filter.register(np.ndarray)
+def numpy_spline_filter(*args, **kwargs):
+    return ndimage.spline_filter
+
+
+@dispatch_spline_filter.register_lazy("cupy")
+def register_cupy_spline_filter():
+    import cupy
+    import cupyx.scipy.ndimage
+
+    @dispatch_affine_transform.register(cupy.ndarray)
+    def cupy_spline_filter(*args, **kwargs):
+
+        return cupyx.scipy.ndimage.spline_filter
+
+
+dispatch_spline_filter1d = Dispatcher(name="dispatch_spline_filter1d")
+
+# ================== spline_filter1d ==================
+@dispatch_spline_filter1d.register(np.ndarray)
+def numpy_spline_filter1d(*args, **kwargs):
+    return ndimage.spline_filter1d
+
+
+@dispatch_spline_filter1d.register_lazy("cupy")
+def register_cupy_spline_filter1d():
+    import cupy
+    import cupyx.scipy.ndimage
+
+    @dispatch_affine_transform.register(cupy.ndarray)
+    def cupy_spline_filter1d(*args, **kwargs):
+
+        return cupyx.scipy.ndimage.spline_filter1d
+
+
 dispatch_asarray = Dispatcher(name="dispatch_asarray")
 
 # ===================== asarray ========================
--- a/dask_image/dispatch/_dispatch_ndinterp.py
+++ b/dask_image/dispatch/_dispatch_ndinterp.py
@@ -43,7 +43,7 @@ def register_cupy_spline_filter():
     import cupy
     import cupyx.scipy.ndimage
 
-    @dispatch_affine_transform.register(cupy.ndarray)
+    @dispatch_spline_filter.register(cupy.ndarray)
     def cupy_spline_filter(*args, **kwargs):
 
         return cupyx.scipy.ndimage.spline_filter
@@ -62,7 +62,7 @@ def register_cupy_spline_filter1d():
     import cupy
     import cupyx.scipy.ndimage
 
-    @dispatch_affine_transform.register(cupy.ndarray)
+    @dispatch_spline_filter1d.register(cupy.ndarray)
     def cupy_spline_filter1d(*args, **kwargs):
 
         return cupyx.scipy.ndimage.spline_filter1d
--- a/dask_image/dispatch/_dispatch_ndinterp.py
+++ b/dask_image/dispatch/_dispatch_ndinterp.py
@@ -13,6 +13,7 @@ __all__ = [
 
 dispatch_affine_transform = Dispatcher(name="dispatch_affine_transform")
 
+
 # ================== affine_transform ==================
 @dispatch_affine_transform.register(np.ndarray)
 def numpy_affine_transform(*args, **kwargs):
@@ -31,6 +32,7 @@ def register_cupy_affine_transform():
 
 
 dispatch_spline_filter = Dispatcher(name="dispatch_spline_filter")
+
 
 # ================== spline_filter ==================
 @dispatch_spline_filter.register(np.ndarray)
@@ -51,6 +53,7 @@ def register_cupy_spline_filter():
 
 dispatch_spline_filter1d = Dispatcher(name="dispatch_spline_filter1d")
 
+
 # ================== spline_filter1d ==================
 @dispatch_spline_filter1d.register(np.ndarray)
 def numpy_spline_filter1d(*args, **kwargs):
@@ -69,6 +72,7 @@ def register_cupy_spline_filter1d():
 
 
 dispatch_asarray = Dispatcher(name="dispatch_asarray")
+
 
 # ===================== asarray ========================
 @dispatch_asarray.register(np.ndarray)
--- a/dask_image/ndinterp/__init__.py
+++ b/dask_image/ndinterp/__init__.py
@@ -115,8 +115,8 @@ def affine_transform(
         raise NotImplementedError(f"Mode {mode} is not currently supported.")
 
     # process kwargs
-    # prefilter is not yet supported
     if prefilter and order > 1:
+        # prefilter is not yet supported for all modes
         if mode in ['nearest', 'grid-constant']:
             raise NotImplementedError(
                 f"order > 1 with mode='{mode}' is not supported."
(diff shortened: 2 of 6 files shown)


grlee77:
cc @m-albert regarding [previous discussion related to this](https://github.com/dask/dask-image/pull/159#pullrequestreview-506964981)

grlee77:
This is a simple performance benchmark on a somewhat large 3D volume (512 x 512 x 512)

from time import time

```Python
import dask.array as da
import numpy as np
import scipy.ndimage as ndi

from dask_image.ndinterp import spline_filter


# define test image
a = 512
ndim = 3
shape = (a,) * ndim
dtype = np.float64
im = np.random.standard_normal(shape).astype(dtype, copy=False)

# transform into dask array
chunksize = (128,) * ndim
dim = da.from_array(im, chunks=chunksize)

kwargs = dict(order=3, mode='constant', output=dtype)

tstart = time()
y_cpu = ndi.spline_filter(im, **kwargs)
dur = time() - tstart
print(f"Duration (scipy.ndimage) = {dur}")

tstart = time()
y = spline_filter(dim, **kwargs)
y = y.compute()
dur = time() - tstart
print(f"Duration (dask-image) = {dur}")
```

On my system with a 10-core Intel Skylake X CPU the result is:

for `order=3`
```
Duration (scipy.ndimage) = 7.603273630142212
Duration (dask-image) = 1.2285776138305664
```
or for `order=5`
```
Duration (scipy.ndimage) = 11.27594804763794
Duration (dask-image) = 1.9986212253570557
```

grlee77:
I see there are quite a few test failures. I think a number of these tests cases will have to be skipped when SciPy <= 1.6. I can fix that locally and then update here, later. 

Currently all cases are passing in my local environment with SciPy 1.6.3 installed.

m-albert:
Hi @grlee77, this is great!

The speedup on the CPU is considerable and of course this allows applying `spline_filter` on the GPU.

Really nice that you added prefilter functionality to `affine_transform`. Actually, the otherwise resulting blur has just been [mentioned](https://github.com/dask/dask-image/issues/210#issue-876254910) by @martinschorb. The implementation and tests for this look really good to me.

> Currently all cases are passing in my local environment with SciPy 1.6.3 installed.

Locally all tests passing for me, too.

>I see there are quite a few test failures. I think a number of these tests cases will have to be skipped when SciPy < 1.6. I can fix that locally and then update here, later.

Looking at the CI tests it seems there could be an error unrelated to  the `scipy` version. In `spline_filter1d`, `map_overlap` seems unable to deal with a too large depth: 

Call of `map_overlap` with:
```python
x = dask.array<array, shape=(48, 48, 48), dtype=float64, chunksize=(33, 33, 33), chunktype=numpy.ndarray>
depth = {0: 0, 1: 0, 2: 22}
boundary = {0: 'none', 1: 'none', 2: 'none'}
```
resulting in:
>E               ValueError: The overlapping depth 22 is larger than your
E               smallest chunk size 15. Rechunk your array
E               with a larger chunk size or a chunk size that
E               more evenly divides the shape of your array.

Probably the affected chunk is the last one along the array, which is smaller than the previous ones. Could this potentially be avoided by previous padding? Strange though that this doesn't occur locally...

grlee77:
> Probably the affected chunk is the last one along the array, which is smaller than the previous ones. Could this potentially be avoided by previous padding? Strange though that this doesn't occur locally...

I am guessing this is related to the Dask version. I have the recent 2021.4.1 release locally, but the CI job I looked at is using 2.8.1 (released Nov 2019).

grlee77:
local testing indicates that the test failures don't occur for dask>=2020.12.0. Specifically, it is this PR that fixed it dask/dask#6708

grlee77:
The one case that is expected to still fail on current dask is if the depth is larger than any of the overall array dimensions. 

I could add a test for that case if desired (e.g. we cannot spline_filter a small image of shape (16, 16) with order=5).

m-albert:
> I am guessing this is related to the Dask version. I have the recent 2021.4.1 release locally, but the CI job I looked at is using 2.8.1 (released Nov 2019).

> I am guessing this is related to the Dask version. I have the recent 2021.4.1 release locally, but the CI job I looked at is using 2.8.1 (released Nov 2019).

Of course, that makes sense! Thanks for checking this.

GenevieveBuckley:
The coveralls check shouldn't block this PR. I've just had a look at it & things there are mostly fine.

dask_image/dispatch/_dispatch_ndinterp.py
```diff
@@ -30,6 +30,44 @@ def cupy_affine_transform(*args, **kwargs):
         return cupyx.scipy.ndimage.affine_transform
 
 
+dispatch_spline_filter = Dispatcher(name="dispatch_spline_filter")
+
```
GenevieveBuckley, line 34: Would you mind running flake8 on the files for this PR? There's a few formatting changes it'll suggest (like here, it will ask you for two blank lines before a function definition).

dask_image/ndinterp/__init__.py
```diff
@@ -105,21 +107,21 @@ def affine_transform(
         offset = matrix[:image.ndim, image.ndim]
         matrix = matrix[:image.ndim, :image.ndim]
 
+    cval = kwargs.pop('cval', 0)
+    mode = kwargs.pop('mode', 'constant')
+    prefilter = kwargs.pop('prefilter', False)
+    if mode in ['wrap', 'reflect', 'mirror', 'grid-mirror', 'grid-wrap']:
+        raise NotImplementedError(f"Mode {mode} is not currently supported.")
```
GenevieveBuckley, line 114: Any chance you can add what modes we *do* support to that error message? That's likely to be a lot less frustrating to users when they run into it. (And yes, I realize you didn't write this message originally, but if it's an easy fix I'd like us to do it)

dask_image/ndinterp/__init__.py
```diff
@@ -105,21 +107,21 @@ def affine_transform(
...
-
-    if 'mode' in kwargs:
-        if kwargs['mode'] in ['wrap', 'reflect', 'mirror']:
-            raise(NotImplementedError("Mode %s is not currently supported."
-                                      % kwargs['mode']))
+    if prefilter and order > 1:
+        # prefilter is not yet supported for all modes
+        if mode in ['nearest', 'grid-constant']:
+            raise NotImplementedError(
+                f"order > 1 with mode='{mode}' is not supported."
```
GenevieveBuckley, line 121: Same here, it would be good to point users in the right direction (i.e. what modes *are* supported that they should try instead). It's nice to have error messages that say both what went wrong and how you might fix it.

dask_image/ndinterp/__init__.py
```diff
@@ -232,3 +234,136 @@ def affine_transform(
                            meta=meta)
 
     return transformed
+
+
+# magnitude of the maximum filter pole for each order
+# (obtained from scipy/ndimage/src/ni_splines.c)
+_maximum_pole = {
```
GenevieveBuckley, line 241: Thank you for the comment explaining the source of these magic numbers :smile:

tests/test_dask_image/test_ndinterp/test_affine_transformation.py
```diff
@@ -1,13 +1,33 @@
...
 # -*- coding: utf-8 -*-
 
+from packaging import version
+
+import dask
+import dask.array as da
+import numpy as np
 import pytest
+import scipy
+from scipy import ndimage
```
GenevieveBuckley, line 11: This is a nitpick, but could we use `import scipy.ndimage` here instead of `from scipy import ndimage`?

I've started to do a bit of code style cleanup here https://github.com/dask/dask-image/pull/227, and `import scipy.ndimage` is what we'll try to use consistently across the codebase.

dask_image/ndinterp/__init__.py
GenevieveBuckley: Looks like 'constant' and 'nearest' are the two safe modes

GenevieveBuckley (approved):
This looks good to me. I've made a few suggestions, and will leave this open for a bit for other reviewers to make comments too.

dask_image/ndinterp/__init__.py
grlee77: yes. also 'grid-constant' if using SciPy 1.6+.

dask_image/dispatch/_dispatch_ndinterp.py
grlee77: done!

dask_image/ndinterp/__init__.py
grlee77: Yeah, given the other wrap, etc. modes that `affine_transform` itself doesn't work with, only 'constant' mode can currently perform the prefiltering.

GenevieveBuckley (approved):

PR #221, opened 2021-05-13
ENH: add remaining kwargs to binary_closing and binary_opening

grlee77:
This PR just adds all currently missing keyword arguments from the SciPy API for `binary_opening` and `binary_closing`. Fortunately, these same arguments were already present (and tested) for the underlying erosion and dilation functions, so this ended up being easy!


GenevieveBuckley (approved):
Well spotted, and good to hear it was an easier than expected fix!

I'm happy with this.

PR #222, opened 2021-05-13
ndfourier: support n > 0 (for rfft) and improve performance

grlee77:
The first commit here support n > 0, which is intended for use when the image to be filtered was created via a real-to-complex FFT. It is mainly just needing to use `dask.array.fft.rfftfreq` instead of `dask.array.fft.fftfreq` on the corresponding axis.

The second commit here refactors the separable filtering to avoid dense frequency arrays of shape `(image.ndim,) + image.shape`. This is done using meshgrid's `sparse=True` option to get a series of separable 1D frequency arrays instead. These can then be applied separably along each axis of the `image`, avoiding the need for `tensordot`. This should reduce computation time and peak memory usage.

I did some local benchmarking and typically saw factor 1.5-3.5 improvement in various cases. This makes sense, as rather than computing, for example, the Gaussian via `dask.array.exp` on the dense frequency grid, we just have to compute it on much smaller 1D arrays instead.

### DIFF
--- a/tests/test_dask_image/test_ndfourier/test_core.py
+++ b/tests/test_dask_image/test_ndfourier/test_core.py
@@ -218,15 +218,28 @@ def test_fourier_filter_non_positive(funcname, s):
         "fourier_uniform",
     ]
 )
-def test_fourier_filter(funcname, s):
+@pytest.mark.parametrize(
+    "n, axis",
+    [
+        ('real', -1),
+        ('real', 0),
+        (-1, -1),
+    ]
+)
+def test_fourier_filter(funcname, s, n, axis):
     da_func = getattr(da_ndf, funcname)
     sp_func = getattr(sp_ndf, funcname)
 
-    a = np.arange(140.0).reshape(10, 14).astype(complex)
+    shape = (10, 14)
+    if n == 'real':
+        n = 2 * shape[axis] - 1
+    dtype = np.float64 if n != -1 else np.complex128
+
+    a = np.arange(140.0).reshape(shape).astype(dtype)
     d = da.from_array(a, chunks=(5, 7))
 
-    r_a = sp_func(a, s)
-    r_d = da_func(d, s)
+    r_a = sp_func(a, s, n=n, axis=axis)
+    r_d = da_func(d, s, n=n, axis=axis)
 
     assert d.chunks == r_d.chunks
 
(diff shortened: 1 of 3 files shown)


grlee77:
@jakirkham, I see that there is `__author__` and `__email__` info for you at the top of the file. Do you want me to also record authorship there or should that be removed?

jakirkham:
Thanks for bring this up! There are a few tasks like this. So just filed a new issue ( https://github.com/dask/dask-image/issues/223 ) about them. Maybe we can handle that in a separate PR?

Agree it would be great to have you on the author list 🙂

tests/test_dask_image/test_ndfourier/test_core.py
```diff
@@ -218,15 +218,28 @@ def test_fourier_filter_non_positive(funcname, s):
         "fourier_uniform",
     ]
 )
-def test_fourier_filter(funcname, s):
+@pytest.mark.parametrize(
+    "n, axis",
+    [
+        ('real', -1),
```
GenevieveBuckley, line 224: It's a bit confusing that n is supposed to be an integer, but here you have it as a string & then later redefine the string to an integer. It might be clearer to have the pytest parameter be a string all the time, and then define n in the body of the test itself.

GenevieveBuckley (approved):
Looks good to me.

The speed improvements you saw sound very encouraging! Thanks for putting effort into this :slightly_smiling_face: 

(Separate to this PR, we should consider adding performance benchmarks for this repository. There's been a few recent PRs targeting performance, and it'd be nice to have a less ad-hoc way for people to benchmark things)

tests/test_dask_image/test_ndfourier/test_core.py
grlee77: good point. I have updated it in adb14ca

grlee77:
> (Separate to this PR, we should consider adding performance benchmarks for this repository. There's been a few recent PRs targeting performance, and it'd be nice to have a less ad-hoc way for people to benchmark things)

I was actually doing this yesterday. I had benchmarked a ton of functions for RAPIDS cuCIM a month or two ago and was able to adapt [these existing cupyx_* files](https://github.com/rapidsai/cucim/tree/branch-0.20/benchmarks/skimage) to add dask-image as a third case alongside SciPy and CuPy. Currently these require having CuPy to run, but it could be modified so that running the GPU cases is not a requirement. [asv](https://asv.readthedocs.io/en/stable/) is the other tool I am familiar with for benchmarking, but I'm not sure it is easy to adapt to the GPU case (I wanted to use `cupyx.time.repeat` to do the timing measurements because it handles device synchronization properly).

GenevieveBuckley:
:tada:

PR #224, opened 2021-05-14
BUG: add missing import of warnings in dask_image.ndmeasure

grlee77:
This PR just adds a missing import of the `warnings` module.

If the user tries to call `dask_image.ndmeasure.sum`, it currently results in a `NameError` being raised rather than the desired warning to use `sum_labels` instead.


GenevieveBuckley:
Good catch, thank you!

Issue #223, opened 2021-05-13
Updating authors


jakirkham:
Currently [this file]( https://github.com/dask/dask-image/blob/main/AUTHORS.rst ) lists only me. However other people have been contributing here. It would be great to have a more comprehensive list.

Also other projects tend to call this `CONTRIBUTORS.txt`. I think `AUTHORS.rst` just came from a Python project template tool. So it might be worth renaming.

Independently some of the source files have `__author__` and `__email__`. The email referenced is dead and as already noted we have more people contributing. We might want to strip these out.

Not sure what other relevant changes we might want to include, but feel free to suggest.

GenevieveBuckley:
1. I absolutely do not want to spend time maintaining an author/contributor list by hand. I thought linking to the contributors page on github would be a good way around that? https://github.com/dask/dask-image/issues/166

2. I don't mind what this file is called. To be honest, I don't mind whether this file exists at all, since this information is easily available at https://github.com/dask/dask-image/graphs/contributors

3. Yes, stripping out `__author__` and `__email__` sounds like a good idea. Even aside from the fact these things go out of date,  the implication an email might be required is a real disincentive to contribute (why would anyone want to hand strangers more avenues to potentially harass them?)

grlee77:
Just relying on the GitHub contributors page is also be fine by me. From the command line, one can run something like `git shortlog --summary --numbered --email` to get similar info.

jakirkham:
> I absolutely do not want to spend time maintaining an author/contributor list by hand. I thought linking to the contributors page on github would be a good way around that?

I thought we could ask people to add themselves. We do this in Zarr for example. That said, I don't feel strongly about it. Some people do like the credit though

GenevieveBuckley:
> I thought we could ask people to add themselves.

That was the original situation here too, but no-one ever updated that file. 

> That said, I don't feel strongly about it. Some people do like the credit though

Sure, some people do like credit. I'm happy to highlight the github contributors page, I think that fills this function. 

I'm more worried about creating only crediting some contributors & not others with a manually updated list. That feels more fraught.

GenevieveBuckley:
Ok, I'll make a PR to remove the `__author__` and `__email__` lines.

Your old email address still appears in the setup.py file, not sure if you want that changed to something else @jakirkham 
https://github.com/dask/dask-image/blob/16687e5ac1f0476bcfa19016aa41a6fa8993dc04/setup.py#L58

jakirkham:
[...]

PR #225, opened 2021-05-14
Remove out of date email address, strip __author__ & __email__

GenevieveBuckley:
This PR removes John's old, out-of-date email address from files in the source code.

> Independently some of the source files have `__author__` and `__email__`. The email referenced is dead and as already noted we have more people contributing. We might want to strip these out.

Closes https://github.com/dask/dask-image/issues/223


dask_image/ndfourier/__init__.py
```diff
@@ -1,12 +1,6 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import division
```
jakirkham, line 3: Unrelated, but we can probably drop this now that the code is Python 3 only. Raised as issue ( https://github.com/dask/dask-image/issues/226 )

jakirkham (approved):

jakirkham:
Thanks Genevieve! 😄

Issue #226, opened 2021-05-14
Drop remaining Python 3isms


jakirkham:
There appear to be some remnants from Python 3 like `from __future__ import`s that could be removed. Would be good to clean these out to simplify the code

PR #227, opened 2021-05-17
Code style cleanup

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/226
Also addresses most of the points in https://github.com/dask/dask-image/issues/101 (except for moving the function definitions out of the `__init__.py` files)

In this PR, I have:

* Removed any `from __future__` imports
* Ran `isort` on all the files in the repository
* Removed the `assert dask` statements from the top of several test modules. (I believe they are superfluous, because if dask is not available there will be an `ImportError` while pytest is collecting tests, so I don't think we'll ever reach that line anyway? If I'm wrong and they're in for a reason, let me know and I'll reverse this change)
* Code style cleanup: Always use `import numpy as np`
* Code style cleanup: Always use `import dask.array as da`
* Code style cleanup: Always use `import scipy.ndimage`
* Code style cleanup: For clarity, use `import dask_image.ndfilters` instead of `import dask_image.ndfilters as da_ndf`
* Code style cleanup: flake8 formatting

### DIFF
--- a/dask_image/ndmeasure/_utils/__init__.py
+++ b/dask_image/ndmeasure/_utils/__init__.py
@@ -1,15 +1,9 @@
 # -*- coding: utf-8 -*-
-
-
-from __future__ import division
-
 import warnings
-
-import numpy
 
 import dask
 import dask.array
-
+import numpy
 
 try:
     from dask.array import blockwise as da_blockwise
--- a/dask_image/ndmeasure/_utils/__init__.py
+++ b/dask_image/ndmeasure/_utils/__init__.py
@@ -3,7 +3,7 @@ import warnings
 
 import dask
 import dask.array
-import numpy
+import numpy as np
 
 try:
     from dask.array import blockwise as da_blockwise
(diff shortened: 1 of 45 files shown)


dask_image/ndmeasure/_utils/__init__.py
```diff
@@ -1,15 +1,9 @@
...
-import numpy
-
 import dask
-import dask.array
-
+import dask.array as da
+import numpy as np
 
 try:
     from dask.array import blockwise as da_blockwise
```
GenevieveBuckley, line 9: @jakirkham is this try statement intended to support a very old version of Dask? Do you think we could get rid of this?

jakirkham: Yeah this was renamed in Dask 1.1.0 ( https://github.com/dask/dask/pull/4348 ). If we are comfortable making that a minimum requirement, we can drop that and just use `dask.array.blockwise` below

GenevieveBuckley: Ok!
I suspect that changing the minimum requirement for Dask is going to involve some fiddling with the environment files we use for the CI, so I'll open an issue & that can be a separate PR.

GenevieveBuckley: Issue made here: https://github.com/dask/dask-image/issues/232

GenevieveBuckley:
If the CI checks all pass, I'll merge this.

PR #240, opened 2021-07-22
Find objects bounding boxes

GenevieveBuckley:
Here's an alternative approach for finding bounding boxes for labelled objects within an image. Closes https://github.com/dask/dask-image/issues/96

This particular implementation would mean you don't need to know the maximum label number ahead of time (an advantage over https://github.com/dask/dask-image/pull/97/).

The approach is roughly as described by @jni [here](https://github.com/dask/dask-image/pull/97/#issuecomment-518256283):
1. For each image chunk, find the bounding boxes for each integer label (except the zero background)
2. Store this information in a dictionary or dataframe, so each integer label is matched to the bounding box.
3. Merge all these together, accounting for the fact that some objects will span multiple image chunks.


GenevieveBuckley:
There are still some flaws I can't quite work out. I'd appreciate thoughts here.

* Problem 1: Right now I need to call compute twice at the end to see the result, so I think I've done something wrong. I've been careful to only delay a single function, so I'm not sure why it's like this. Adding pdb statements isn't helping me drop into problem areas to check.
* Problem 2: I think we should be able to have the dataframes be Dask dataframes instead of pandas dataframes. But when I change `_find_bounding_boxes` to return a Dask dataframe with one partition, then later functions complain about not knowing what meta should be & my attempts to add it have not been successful. Expand the details below to see the error message.

<details>
<summary>Details:</summary>

```python-traceback
distributed.worker - WARNING - Compute Failed
Function:  combine
args:      (Dask DataFrame Structure:
                    0       1
npartitions=1                
111            object  object
222               ...     ...
Dask Name: from_pandas, 1 tasks, Dask DataFrame Structure:
                    0       1
npartitions=1                
222            object  object
333               ...     ...
Dask Name: from_pandas, 1 tasks, <function _combine_dataframes at 0x7fb7c8688e50>)
kwargs:    {}
Exception: ValueError('Metadata inference failed in `combine`.\n\nYou have supplied a custom function and Dask is unable to \ndetermine the type of output that that function returns. \n\nTo resolve this please provide a meta= keyword.\nThe docstring of the Dask function you ran should have more information.\n\nOriginal error is below:\n------------------------\nAttributeError("\'str\' object has no attribute \'start\'")\n\nTraceback:\n---------\n  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/utils.py", line 176, in raise_on_meta_error\n    yield\n  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py", line 5612, in _emulate\n    return func(*_extract_meta(args, True), **_extract_meta(kwargs, True))\n  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/utils.py", line 963, in __call__\n    return getattr(obj, self.method)(*args, **kwargs)\n  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/pandas/core/frame.py", line 6383, in combine\n    arr = func(series, otherSeries)\n  File "/tmp/ipykernel_283799/2299891482.py", line 54, in _combine_dataframes\n  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/pandas/core/series.py", line 2929, in combine\n    new_values.append(func(lv, rv))\n  File "/tmp/ipykernel_283799/2299891482.py", line 48, in _combine_series\n')

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/utils.py in raise_on_meta_error()
    175     try:
--> 176         yield
    177     except Exception as e:

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py in _emulate()
   5611     with raise_on_meta_error(funcname(func), udf=kwargs.pop("udf", False)):
-> 5612         return func(*_extract_meta(args, True), **_extract_meta(kwargs, True))
   5613 

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/utils.py in __call__()
    962     def __call__(self, obj, *args, **kwargs):
--> 963         return getattr(obj, self.method)(*args, **kwargs)
    964 

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/pandas/core/frame.py in combine()
   6382 
-> 6383             arr = func(series, otherSeries)
   6384             arr = maybe_downcast_to_dtype(arr, new_dtype)

/tmp/ipykernel_283799/2299891482.py in _combine_dataframes()
     53 def _combine_dataframes(s1, s2):
---> 54     combined = s1.combine(s2, _combine_series)
     55     return combined

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/pandas/core/series.py in combine()
   2928                 with np.errstate(all="ignore"):
-> 2929                     new_values.append(func(lv, rv))
   2930         else:

/tmp/ipykernel_283799/2299891482.py in _combine_series()
     47     else:
---> 48         start = min(a.start, b.start)
     49         stop = max(a.stop, b.stop)

AttributeError: 'str' object has no attribute 'start'

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
/tmp/ipykernel_283799/2906121027.py in <module>
----> 1 result.compute().compute()

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/base.py in compute(self, **kwargs)
    283         dask.base.compute
    284         """
--> 285         (result,) = compute(self, traverse=False, **kwargs)
    286         return result
    287 

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/base.py in compute(*args, **kwargs)
    565         postcomputes.append(x.__dask_postcompute__())
    566 
--> 567     results = schedule(dsk, keys, **kwargs)
    568     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    569 

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/distributed/client.py in get(self, dsk, keys, workers, allow_other_workers, resources, sync, asynchronous, direct, retries, priority, fifo_timeout, actors, **kwargs)
   2705                     should_rejoin = False
   2706             try:
-> 2707                 results = self.gather(packed, asynchronous=asynchronous, direct=direct)
   2708             finally:
   2709                 for f in futures.values():

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/distributed/client.py in gather(self, futures, errors, direct, asynchronous)
   2019             else:
   2020                 local_worker = None
-> 2021             return self.sync(
   2022                 self._gather,
   2023                 futures,

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/distributed/client.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    860             return future
    861         else:
--> 862             return sync(
    863                 self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
    864             )

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    336     if error[0]:
    337         typ, exc, tb = error[0]
--> 338         raise exc.with_traceback(tb)
    339     else:
    340         return result[0]

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/distributed/utils.py in f()
    319             if callback_timeout is not None:
    320                 future = asyncio.wait_for(future, callback_timeout)
--> 321             result[0] = yield future
    322         except Exception:
    323             error[0] = sys.exc_info()

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/tornado/gen.py in run(self)
    760 
    761                     try:
--> 762                         value = future.result()
    763                     except Exception:
    764                         exc_info = sys.exc_info()

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/distributed/client.py in _gather(self, futures, errors, direct, local_worker)
   1884                             exc = CancelledError(key)
   1885                         else:
-> 1886                             raise exception.with_traceback(traceback)
   1887                         raise exc
   1888                     if errors == "skip":

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/utils.py in __call__()
    961 
    962     def __call__(self, obj, *args, **kwargs):
--> 963         return getattr(obj, self.method)(*args, **kwargs)
    964 
    965     def __reduce__(self):

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py in combine()
   2786     @derived_from(pd.DataFrame)
   2787     def combine(self, other, func, fill_value=None, overwrite=True):
-> 2788         return self.map_partitions(
   2789             M.combine, other, func, fill_value=fill_value, overwrite=overwrite
   2790         )

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py in map_partitions()
    689         None as the division.
    690         """
--> 691         return map_partitions(func, self, *args, **kwargs)
    692 
    693     @insert_meta_param_description(pad=12)

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py in map_partitions()
   5666         # Use non-normalized kwargs here, as we want the real values (not
   5667         # delayed values)
-> 5668         meta = _emulate(func, *args, udf=True, **kwargs)
   5669     else:
   5670         meta = make_meta(meta, index=meta_index, parent_meta=parent_meta)

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py in _emulate()
   5610     """
   5611     with raise_on_meta_error(funcname(func), udf=kwargs.pop("udf", False)):
-> 5612         return func(*_extract_meta(args, True), **_extract_meta(kwargs, True))
   5613 
   5614 

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/contextlib.py in __exit__()
    129                 value = type()
    130             try:
--> 131                 self.gen.throw(type, value, traceback)
    132             except StopIteration as exc:
    133                 # Suppress StopIteration *unless* it's the same exception that

~/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/utils.py in raise_on_meta_error()
    195         )
    196         msg = msg.format(" in `{0}`".format(funcname) if funcname else "", repr(e), tb)
--> 197         raise ValueError(msg) from e
    198 
    199 

ValueError: Metadata inference failed in `combine`.

You have supplied a custom function and Dask is unable to 
determine the type of output that that function returns. 

To resolve this please provide a meta= keyword.
The docstring of the Dask function you ran should have more information.

Original error is below:
------------------------
AttributeError("'str' object has no attribute 'start'")

Traceback:
---------
  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/utils.py", line 176, in raise_on_meta_error
    yield
  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/dataframe/core.py", line 5612, in _emulate
    return func(*_extract_meta(args, True), **_extract_meta(kwargs, True))
  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/dask/utils.py", line 963, in __call__
    return getattr(obj, self.method)(*args, **kwargs)
  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/pandas/core/frame.py", line 6383, in combine
    arr = func(series, otherSeries)
  File "/tmp/ipykernel_283799/2299891482.py", line 54, in _combine_dataframes
  File "/home/genevieve/anaconda3/envs/dask-image-dev-moderndask/lib/python3.8/site-packages/pandas/core/series.py", line 2929, in combine
    new_values.append(func(lv, rv))
  File "/tmp/ipykernel_283799/2299891482.py", line 48, in _combine_series


```

</details>

GenevieveBuckley:
Some other points (not problems, just things to note)

I think it's unavoidable that we'll have to use delayed. 

We can't use the block_info keyword argument directly if map_blocks returns a non-array type of output. We want to return a dictionary or dataframe. See details at https://github.com/dask/dask/issues/7921. Additionally, even if we *do* fix this issue, then we need a more modern version of dask to get map_blocks to return dataframe output. But upgrading the version of dask (eg: to 2021.6.2) causes several other tests to fail in ndmeasure.

Second, I think it's better to avoid using the `scipy.ndimage.find_objects` function directly. If you have an image chunk with just one object with a really high integer label `n`, the scipy `find_objects` result will return `n - 1` values of None, and then the single meaningful result. That seems bad for parallized applications, so I think looping through only the unique integer values present in a given image chunk is a better way to go.

tests/test_dask_image/test_ndmeasure/test_find_objects.py
```diff
@@ -0,0 +1,36 @@
...
+    label_image = np.zeros((5, 10)).astype(int)
+    label_image[1:3,0:2] = 111
+    label_image[3,3:-2] = 222
+    label_image[0:2,-3:] = 333
+    label_image = da.from_array(label_image, chunks=(5, 5))
+    return label_image
+
+
+def test_bounding_boxes(label_image):
+    result = dask_image.ndmeasure.find_objects(label_image).compute().compute()
```
GenevieveBuckley, line 31: This is very irritating: I don't know why I need to call compute twice here, or how to fix that.

GenevieveBuckley: Here's a copy of what I wrote in the dask slack about this problem:

> After running the fold operation, I get a dask.bag.core.Item as the result. If you call compute, you get a dask dataframe, but then you need to call compute again to, you know, compute the actual result. How can I avoid users having to call compute twice?Things I tried:
> * Putting one of the compute statements inside my own function. This isn't great, because it runs a bunch of the computation immediately.
> * Turning the dask.bag.core.Item into a dask dataframe. 
>    * There is a to_dataframe method for regular bags, but this is not available for dask bag Items.
>    * Dask bag Items do have a to_delayed method, so I could do that & then use the dask dataframe dd.from_delayed function. This would work great, except that Dask is doing some type checking and raising an error "Expected partition of type `pandas.core.frame.DataFrame` but got `dask.dataframe.core.DataFrame`". I've been trying to solve that by using the parent_meta kwarg in the dask.dataframe.utils.make_meta function - but it can't dispatch to a dask dataframe backend (only pandas and sparse backends are supported).
> 
> My current plan is to try creating a metadata dataframe without using the convenience make_meta function, and see how that goes.

> Update: nope, that didn't seem to help. I tried wrapping the pandas metadata object inside a dask dataframe, but the type mismatch error is still raised
> `meta = dd.from_pandas(dd.utils.make_meta([(i, object) for i in range(label_image.ndim)]), npartitions=1)`

> Oh, maybe I have it!
> In `dd.from_delayed` there is a `verify_meta` keyword argument, and I can set that to False to skip the partition type checking. 

> For reasons I don't understand, it still seems like you need to call compute twice on the output. Argh
> It's late, I'm going to leave this until Monday. For anyone who'd like more context, the PR I've been working on is here: https://github.com/dask/dask-image/pull/240

dask_image/ndmeasure/_utils/_find_objects.py
```diff
@@ -0,0 +1,57 @@
...
+    data = {}
+    for i in range(ndim):
+        slices = [x[ii] for ii in x.index if str(ii).startswith(str(i))]
+        combined_slices = _combine_slices(slices)
+        data[i] = combined_slices
+    result = pd.Series(data=data, index=[i for i in range(ndim)], name=x.name)
+    return result
+
+
+def _find_objects(df1, df2, ndim=2):
```
GenevieveBuckley, line 49: Still need to figure out how I can pass in a keyword argument for either `ndim`/`meta` (probably more efficient to just pass `meta` in rather than continually recompute it). It doesn't seem like this is possible to do with dask bag fold?

I might be able to figure out a workaround by counting all the dataframe column headings with some regex, but I'd prefer not to do that.

tests/test_dask_image/test_ndmeasure/test_find_objects.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    label_image = np.zeros((5, 10)).astype(int)
+    label_image[1:3,0:2] = 111
+    label_image[3,3:-2] = 222
+    label_image[0:2,-3:] = 333
+    label_image = da.from_array(label_image, chunks=(5, 5))
+    return label_image
+
+
+def test_bounding_boxes(label_image):
```
GenevieveBuckley, line 30: Reminder: this test only passes because it's a 2D label image. Figure out how to pass meta in to the fold function to fix this properly.

GenevieveBuckley:
Here's the task graph for a small example (not the one in the tests, but a small image of blobs with 18 objects spread over four image chunks).

image

GenevieveBuckley:
And after the first compute call (why do I need to call compute twice, what am I doing wrong), the task graph for the same example looks like this:

image

tests/test_dask_image/test_ndmeasure/test_find_objects.py
```diff
@@ -0,0 +1,36 @@
...
+    result = dask_image.ndmeasure.find_objects(label_image).compute().compute()
+    expected = pd.DataFrame.from_dict(
+        {0: {111: slice(1, 3, 1), 222: slice(3, 4, 1), 333: slice(0, 2, 1)},
+        1: {111: slice(0, 2, 1), 222: slice(3, 8, 1), 333: slice(7, 10, 1)}}
```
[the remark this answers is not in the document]
m-albert: When trying out different input arrays

```python
label_image = np.random.randint(0,100, (3, 3))
label_image = ndimage.zoom(label_image, zoom=3, order=1)
label_image = ndimage.label(label_image>50)[0].astype(int)
label_image = da.from_array(label_image, chunks=(5, 5))
```

I often get (not always, I think when there's at least one merge):
> ValueError: Metadata mismatch found in `from_delayed`.

dask_image/ndmeasure/_utils/_find_objects.py
```diff
@@ -0,0 +1,57 @@
...
+        start = min([sl.start for sl in slices])
+        stop = max([sl.stop for sl in slices])
+        return slice(start, stop, 1)
+
+
+def _merge_bounding_boxes(x, ndim):
+    x = x.dropna()
+    data = {}
+    for i in range(ndim):
+        slices = [x[ii] for ii in x.index if str(ii).startswith(str(i))]
```
m-albert, line 42: I'm a bit confused about this line as it mixes the dimension with the label... maybe an explanatory comment would be useful here?

tests/test_dask_image/test_ndmeasure/test_find_objects.py
m-albert: I think problems can occur when mixing the delayed framework with dask arrays.

Replacing

```python
bag = db.from_sequence(arrays)
```
by
```python
    from dask import delayed, compute
    bag = db.from_delayed([delayed(compute)(ar) for ar in arrays])
```
makes it compute after a single call 😅. However this apart from being hacky might come with suboptimal graph optimization?

m-albert (commented):
Hey Genevieve, I think this is great functionality for `dask-image`. I left you some comments in review boxes :)

dask_image/ndmeasure/_utils/_find_objects.py
GenevieveBuckley: Martin has also suggested I try using functools partial, which sounds like a good approach

dask_image/ndmeasure/_utils/_find_objects.py
GenevieveBuckley: Hm, yes this is confusing. Basically, the column naming convention is `0_x, 1_x, 2_x, ... 0_y, 1_y, 2_y, ...`. The number before the underscore indicates which array dimension it belongs to (and the `x`, `y`, etc. are how different origin dataframes are distringuished in the merge with an outer join). We need to find all the columns starting with the same number (array dimension), and then for each row (label) we find the union of those slices.

pseudocode explanation
```
for each array dimension:
    for each label, pick out the values belonging to that array dimension
        and combine those slices
```

I'm not sure I've explained this well, so asking for further clarification or suggesting other improvements would be welcome.

tests/test_dask_image/test_ndmeasure/test_find_objects.py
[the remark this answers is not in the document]
GenevieveBuckley: > Just saw that ndimage.find_objects returns slices such as `slice(1,3,None)` (instead of `slice(1,3,1)`)

Ah, that looks like my mistake. I had assumed that the default slice step was 1, but instead the default value should be `None`. So I think we're seeing None anytime the label doesn't cross the border of an array chunk, because I didn't need to alter/combine slice information.

I don't think this particular thing will have made a material difference, but it's easy for me to fix what I'm doing to match a default step size of None.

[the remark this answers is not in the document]
GenevieveBuckley: The metadata mismatch problem I still need to fix (thanks for pointing it out!)

tests/test_dask_image/test_ndmeasure/test_find_objects.py
GenevieveBuckley: For a basic example it seems nicely parallelized. I think we'd need it to be *very* suboptimal before I'd be happy about telling users to call compute twice, it's not a clean experience for them.

image

GenevieveBuckley: Thanks a bunch for suggesting this @m-albert 

I was unhappy about this issue, and I think you've solved it nicely

GenevieveBuckley: (Turns out we want to put this delayed compute call a bit later on, just before we tell it the result should be a dask dataframe. Works great!)

tests/test_dask_image/test_ndmeasure/test_find_objects.py
```diff
@@ -0,0 +1,40 @@
...
+import dask.dataframe as dd
+import numpy as np
+import pandas as pd
+import pytest
+
+import dask_image.ndmeasure
+
+
+@pytest.fixture
+def label_image():
```
GenevieveBuckley, line 11: Separately, it would also be good to construct a really large test example (not to keep/run in the tests, just to give this thing a whirl and see what it can do)

dask_image/ndmeasure/_utils/_find_objects.py
GenevieveBuckley: Martin's suggestion of functools partial is working really well :)

tests/test_dask_image/test_ndmeasure/test_find_objects.py
GenevieveBuckley: Done

dask_image/ndmeasure/_utils/_find_objects.py
```diff
@@ -0,0 +1,57 @@
...
+    for idx, chunk in zip(block_id, chunks):
+        array_location.append(sum(chunk[:idx]))
+    return tuple(array_location)
+
+
+@delayed
+def _find_bounding_boxes(x, array_location):
+    """An alternative to scipy.ndi.find_objects"""
+    unique_vals = np.unique(x)
+    unique_vals = unique_vals[unique_vals != 0]
```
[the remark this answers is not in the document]
GenevieveBuckley: We could potentially do that, as long as we kept track of the mapping between the old and new label integers.

Whether it's faster & worth it would depend on results from some performance testing. I'm inclined to get an implementation in, and then tinker with speed improvements (and anyone who'd like to jump in and try stuff is more than welcome!)

tests/test_dask_image/test_ndmeasure/test_find_objects.py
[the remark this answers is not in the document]
GenevieveBuckley: I'm not able to replicate your metadata mismatch error, @m-albert. Do you have a minimal, reproducible example?

dask_image/ndmeasure/_utils/_find_objects.py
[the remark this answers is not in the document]
m-albert: Sounds good :)

tests/test_dask_image/test_ndmeasure/test_find_objects.py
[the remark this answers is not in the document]
m-albert: I'm getting the error with this code using the latest dask and PR branch.
```python
import numpy as np
import dask.array as da
import dask_image.ndmeasure
from scipy import ndimage

np.random.seed(0)
label_image = np.random.randint(0,100, (3, 3))
label_image = ndimage.zoom(label_image, zoom=3, order=1)
label_image = ndimage.label(label_image>50)[0].astype(int)
label_image = da.from_array(label_image, chunks=(5, 5))

result = dask_image.ndmeasure.find_objects(label_image).compute()
```
With `np.random.seed(1)` instead it runs through

dask_image/ndmeasure/_utils/_find_objects.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _merge_bounding_boxes(x, ndim):
+    """Merge the bounding boxes describing objects over multiple image chunks."""
+    x = x.dropna()
+    data = {}
+    # For each dimension in the array,
+    # pick out the slice values belonging to that dimension
+    # and combine those slices (find the union; the slice expanded to all input slices).
+    for i in range(ndim):
+        # Array dimensions are labelled by a number followed by an underscroe
```
m-albert, line 54: Suggestion for further comment line: "x and y represent the pair of chunks from which label slices are merged"

dask_image/ndmeasure/__init__.py
```diff
@@ -202,6 +206,37 @@ def extrema(image, label_image=None, index=None):
     return result
 
 
+def find_objects(label_image):
```
m-albert, line 209: Suggestion: indicate here that a dask dataframe is returned instead of a list as in the ndimage version

GenevieveBuckley:
I'm hoping https://github.com/dask/dask/issues/7851 isn't going to be a problem here (might not be, but it's a good idea to try this on something of a decent size)

tests/test_dask_image/test_ndmeasure/test_find_objects.py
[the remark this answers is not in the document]
GenevieveBuckley: Hm, I am able to replicate this bug today. I'll have another look at it now.

GenevieveBuckley:
I'd forgotten we'd left this one hanging. Reviewing the thread, it seems like [this](https://github.com/dask/dask-image/pull/240#discussion_r681064071) is the blocker.

dask_image/ndmeasure/__init__.py
GenevieveBuckley: Good suggestion, done.

dask_image/ndmeasure/_utils/_find_objects.py
GenevieveBuckley: Added, thank you.

dask_image/ndmeasure/__init__.py
```diff
@@ -202,6 +206,37 @@ def extrema(image, label_image=None, index=None):
...
+
+    arrays = []
+    for block_id, block in block_iter:
+        array_location = _array_chunk_location(block_id, label_image.chunks)
+        arrays.append(delayed(_find_bounding_boxes)(block, array_location))
+
+    bag = db.from_sequence(arrays)
+    result = bag.fold(functools.partial(_find_objects, label_image.ndim), split_every=2).to_delayed()
+    meta = dd.utils.make_meta([(i, object) for i in range(label_image.ndim)])
+    result = delayed(compute)(result)[0]  # avoid the user having to call compute twice on result
```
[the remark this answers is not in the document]
GenevieveBuckley: This is an interesting point. I am inclined to leave this for a follow up PR.

There are lots of ways to specify which scheduler, etc. and passing them into the compute call is only one way. So even if this does not support it, users can still do things like use the context manager..

We should make this into an issue, so the idea doesn't get lost and forgotten.

tests/test_dask_image/test_ndmeasure/test_find_objects.py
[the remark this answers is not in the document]
GenevieveBuckley: Ok, it seems the problem happens when at least one of the label image chunks has no non-zero labels in it. 

That means there is an empty dask array within the `arrays` list, which then doesn't play well with the `_find_objects` function (I think it's the `dd.merge(df1, df2, ...` line that's the problem, still investigating).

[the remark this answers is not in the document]
GenevieveBuckley: Update: the problem is actually with `_find_objects` in this line `df1 = dd.from_delayed(df1, meta=meta)`. If df1 is a delayed object of an empty dask dataframe, we get the metadata mismatch error.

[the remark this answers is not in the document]
GenevieveBuckley: Found and fixed!

Turns out this was easy to fix - I just had to make sure that the pandas dataframe we create in `_find_bounding_boxes` from a dictionary always has the column names specified. That way if the dictionary is empty (no non-zero labels found in that chunk), we still get the right shape of dataframe.

dask_image/ndmeasure/__init__.py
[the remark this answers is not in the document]
GenevieveBuckley: Issue made here: https://github.com/dask/dask-image/issues/251

tests/test_dask_image/test_ndmeasure/test_find_objects.py
GenevieveBuckley: Going to release this into the wild and see if anyone out there will end up road testing this for us (I've left it languish here too long while I've been meaning to get around to it)

dask_image/ndmeasure/_utils/_find_objects.py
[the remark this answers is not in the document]
GenevieveBuckley: Issue made to track any future performance testing of `find_objects` https://github.com/dask/dask-image/issues/252

tests/test_dask_image/test_ndmeasure/test_find_objects.py
GenevieveBuckley: Issue made here to track any future performance testing of `find_objects` https://github.com/dask/dask-image/issues/252

GenevieveBuckley:
Found and fixed the bug holding us up. Will merge today.

GenevieveBuckley:
Final note: this new function is not compatible with cupy. That'll be a thing to work on later, see https://github.com/dask/dask-image/issues/253

PR #241, opened 2021-07-29
Upgrade for compatibility with latest dask release

GenevieveBuckley:
The small update in this PR means our tests will pass with both old and newer dask versions.

The version of dask we have pinned in the test environments is quite old (as in, before the calver standard was adopted for dask releases). I want to depend on a more up-to-date version of dask (eg: for https://github.com/dask/dask-image/pull/240).


jakirkham (approved):

jakirkham:
Thanks Genevieve! 😄

Issue #242, opened 2021-09-21
ndfilters.convolve wrap mode not  working when the array is chunked


anlavandier:
**What happened**: 
Used `ndfilters.convolve` using the `wrap` mode and the output had the wrong values on the edges
**What you expected to happen**:
The function to output the correct values on the edges. 
**Minimal Complete Verifiable Example**:

```python
import numpy as np
import dask.array as da
from dask_image.ndfilters import convolve


def test_correct(F,axis = 0):
    if axis !=0:
        F = da.swapaxes(F, 0, axis)
        result = test_correct(F,axis = 0)
        return result
    
    kernel = np.array([1,-1,0,1,-1]).reshape((5,) + (1,) * (F.ndim - 1))
    # The output using ndfilters.convolve with the wrap mode
    O1 = convolve(F,kernel, mode = 'wrap')
    # Output created by forcing the wrapping using dask.array.overlap.overlap then calculating the convolution. 
    # taking care of the boundary
    F_1 = da.overlap.overlap(F, depth = {axis : 2}, boundary = 'periodic')
    # calculate the convolution
    O2_int = convolve(F_1,kernel, mode = 'constant')
    # removing the useless ghosts zones. 
    O2 = da.overlap.trim_overlap(O2_int, depth = {axis : 2})

    return O1.compute(), O2.compute()

test_1D_no_chunks = da.from_array([1,2,3,4,5,6,7,8,9],chunks = (-1,))
test_1D_chunked = da.from_array([1,2,3,4,5,6,7,8,9],chunks = (5,))

print('1D no chunks :',*test_correct(test_1D_no_chunks), sep = '\n', end = '\n\n')
print('1D chunked :', *test_correct(test_1D_chunked), sep = '\n', end = '\n\n')

test_2D_no_chunks = da.from_array(np.arange(25).reshape(5,5),chunks =(-1,-1))
test_2D_chunked = da.from_array(np.arange(25).reshape(5,5),chunks =(-1,2))

print('2D no chunks axis 0  :',*test_correct(test_2D_no_chunks,axis = 0), sep = '\n\n', end = '\n\n\n')
print('2D no chunks axis 1  :',*test_correct(test_2D_no_chunks,axis = 1), sep = '\n\n', end = '\n\n\n')

print('2D chunked axis 0  :',*test_correct(test_2D_chunked,axis = 0), sep = '\n\n', end = '\n\n\n')
[...]

PR #243, opened 2021-09-24
Fix wrap bug in ndfilters convolve and correlate

anlavandier:
Closes [#242]
This bugged was caused by letting Scipy handle the boundary conditions which cannot work when `mode = "wrap"` and the array has more than one chunk. 
Quick explanation of what happened : 
The old function does 
```python
   # mode is untouched from the call to ndfilters.convolve, boundary == "none" and everything else is irrelevant
    result = image.map_overlap(
        dispatch_convolve(image),
        depth=depth,
        boundary=boundary,
        dtype=image.dtype,
        meta=image._meta,
        weights=weights,
        mode=mode,
        cval=cval,
        origin=origin,
    )
```
This means that we will apply `scipy.ndimage.convolve` to each chunk. When `mode == "wrap"`, the wrapping will be done on the chunks which means that the values used for the boundaries are the edge values of the chunk and not the array, hence why it doesn't work properly. This is the only mode for which it happens because wrapping is the only case where being limited to a chunk is a problem. 

The fix I used is just those three lines. 
```python
    if mode == "wrap":
        boundary = "periodic"
        mode = "constant"
```

This also applies to `ndfilters.correlate`.

### DIFF
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -47,7 +47,7 @@ def correlate(image, weights, mode="reflect", cval=0.0, origin=0):
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
-    if mode == "wrap":
+    if mode == "wrap":  # Fixes https://github.com/dask/dask-image/issues/242
         boundary = "periodic"
         mode = "constant"
 
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -20,7 +20,7 @@ def convolve(image, weights, mode="reflect", cval=0.0, origin=0):
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
-    if mode == "wrap":
+    if mode == "wrap":  # Fixes https://github.com/dask/dask-image/issues/242
         boundary = "periodic"
         mode = "constant"
 
--- a/dask_image/ndfilters/_conv.py
+++ b/dask_image/ndfilters/_conv.py
@@ -13,16 +13,16 @@ __all__ = [
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.convolve)
-def convolve(image,
-             weights,
-             mode='reflect',
-             cval=0.0,
-             origin=0):
+def convolve(image, weights, mode="reflect", cval=0.0, origin=0):
     check_arraytypes_compatible(image, weights)
 
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
+
+    if mode == "wrap":
+        boundary = "periodic"
+        mode = "constant"
 
     result = image.map_overlap(
         dispatch_convolve(image),
@@ -33,23 +33,23 @@ def convolve(image,
         weights=weights,
         mode=mode,
         cval=cval,
-        origin=origin
+        origin=origin,
     )
 
     return result
 
 
 @_utils._update_wrapper(scipy.ndimage.filters.correlate)
-def correlate(image,
-              weights,
-              mode='reflect',
-              cval=0.0,
-              origin=0):
+def correlate(image, weights, mode="reflect", cval=0.0, origin=0):
     check_arraytypes_compatible(image, weights)
 
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
+
+    if mode == "wrap":
+        boundary = "periodic"
+        mode = "constant"
 
     result = image.map_overlap(
         dispatch_correlate(image),
@@ -60,7 +60,7 @@ def correlate(image,
         weights=weights,
         mode=mode,
         cval=cval,
-        origin=origin
+        origin=origin,
     )
 
     return result
--- a/tests/test_dask_image/test_ndfilters/test__conv.py
+++ b/tests/test_dask_image/test_ndfilters/test__conv.py
@@ -154,3 +154,35 @@ def test_convolutions_compare(sp_func,
             d, weights, origin=origin
         )
     )
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


anlavandier:
Sidenote: I was not able to properly run py.test because it complained about not being able to import dask_image. Nonetheless, I was done testing on my own and there is not reason for the test that I added not to work.

jakirkham:
Ah ok. This is part of what I was wondering about in issue. Namely if something might be going wrong in SciPy or perhaps some handling around it. Thanks for digging into this! 😄

Have started CI so we can get a sense of where things stand on the testing front and hopefully go from there 🙂

dask_image/ndfilters/_conv.py
```diff
@@ -13,17 +13,17 @@
...
-             cval=0.0,
-             origin=0):
+def convolve(image, weights, mode="reflect", cval=0.0, origin=0):
     check_arraytypes_compatible(image, weights)
 
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
+    if mode == "wrap":
```
GenevieveBuckley, line 23:
```suggestion
    if mode == "wrap":  # Fixes https://github.com/dask/dask-image/issues/242
```

The reason is clear now, but future developers may wonder why this is a special case. Adding a note will save them time later.

tests/test_dask_image/test_ndfilters/test__conv.py
```diff
@@ -154,3 +154,35 @@ def test_convolutions_compare(sp_func,
             d, weights, origin=origin
         )
     )
+@pytest.mark.parametrize(
```
GenevieveBuckley, line 157: I love how comprehensive this test is with pytest parameterize. Unfortunately this test passes *both* before and after your change, so it's not a good test. But this is easy to fix - in the original bug report we only saw the bug for 1D arrays, so you should choose a 1D array example here as well.

dask_image/ndfilters/_conv.py
```diff
@@ -13,17 +13,17 @@
...
+        boundary = "periodic"
+        mode = "constant"
```
GenevieveBuckley, line 25: I don't think it's necessary to include this line? 

When I comment it out (effectively leaving behind `mode="reflect"`) I don't see a difference in the final result. I have only looked at very simple examples though, so if you can explain it more that would be helpful.

dask_image/ndfilters/_conv.py
```diff
@@ -33,24 +33,24 @@ def convolve(image,
...
+def correlate(image, weights, mode="reflect", cval=0.0, origin=0):
     check_arraytypes_compatible(image, weights)
 
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
+    if mode == "wrap":
+        boundary = "periodic"
+        mode = "constant"
```
GenevieveBuckley, line 52: Same as above

tests/test_dask_image/test_ndfilters/test__conv.py
GenevieveBuckley, line 157: Love how thorough this test is using pytest parametrize!

(For a brief moment I thought this test might need a 1D instead of 2D example, but on reflection I don't think that matters)

dask_image/ndfilters/_conv.py
```diff
@@ -33,24 +33,24 @@ def convolve(image,
...
-              cval=0.0,
-              origin=0):
+def correlate(image, weights, mode="reflect", cval=0.0, origin=0):
     check_arraytypes_compatible(image, weights)
 
     origin = _utils._get_origin(weights.shape, origin)
     depth = _utils._get_depth(weights.shape, origin)
     depth, boundary = _utils._get_depth_boundary(image.ndim, depth, "none")
 
+    if mode == "wrap":
```
GenevieveBuckley, line 50:
```suggestion
    if mode == "wrap":  # Fixes https://github.com/dask/dask-image/issues/242
```

GenevieveBuckley:
Thank you for the pull request @anlavandier

dask_image/ndfilters/_conv.py
anlavandier: It is not necessary at all indeed. The reason I did it is because I assume that  it is (barely) faster than doing reflect because padding with constant values is easier than padding reflectively.

tests/test_dask_image/test_ndfilters/test__conv.py
anlavandier: I saw that you previously said that the test works even without my changes. I'm not sure that I understand how it is possible. I tested myself on the exact same values than in the test and `mode = "wrap"` shows a discrepancy between `scipy.ndimage` and `dask_image` that dissappears after adding the changes I made.

dask_image/ndfilters/_conv.py
GenevieveBuckley: Ah, I see, you're probably right about that. 
As a general rule, I think that making changes for performance should happen after a benchmarking script shows an improvement.

GenevieveBuckley:
@jakirkham do you know why we have to keep clicking "approve" so the CI runs on this PR? I had assumed once it was clicked once, then all subsequent commits on the PR would run on the CI, but that doesn't seem to be happening. If there's a setting we should change somewhere, maybe that could help?

tests/test_dask_image/test_ndfilters/test__conv.py
GenevieveBuckley: I was trying to do two things at once and made a silly mistake earlier. It's ok!

GenevieveBuckley:
CI checks all look good, I'll merge this now. Ideally we'll get a new dask-image release out in the next few weeks (momentum here is a bit stalled because I'd like to do some benchmarking for imread before that happens). We'll see how that goes.

Thank you for contributing @anlavandier! Let us know if there's anything else you think would be useful to have.

jakirkham:
> @jakirkham do you know why we have to keep clicking "approve" so the CI runs on this PR? I had assumed once it was clicked once, then all subsequent commits on the PR would run on the CI, but that doesn't seem to be happening. If there's a setting we should change somewhere, maybe that could help?

Yeah that's how GitHub implemented this ( https://github.com/dask/community/issues/158 ). Not much we can do about it. That said, users who have committed to the repo shouldn't still need approving

PR #244, opened 2021-10-04
Convolution based on scipy.signal's function

anlavandier:
This pull request is the result of another pull request on the main dask repository:  https://github.com/dask/dask/pull/8149

The convolution function has been added in the newly created dask_image.signal. Appropriate tests (that I couldn't run on my computer for some reason) have also been added in `test_dask_image/test_signal/`. There is also a docstring for the function though I don't understand why dask and dask_image build different HTML pages with the exact same docstring. 

Finally, while I was directed here by the core dask team, I personally think that this is even more confusing than adding the convolution in `dask.array` because it has basically nothing to do with `scipy.ndimage` or image processing in general.

### DIFF
--- a/continuous_integration/environment-3.6.yml
+++ b/continuous_integration/environment-3.6.yml
@@ -13,7 +13,7 @@ dependencies:
   - pytest-flake8==1.0.4
   - dask==2.8.1
   - numpy==1.15.4
-  - scipy==1.2.0
+  - scipy==1.4.0
   - scikit-image==0.14.2
   - pims==0.4.1
   - slicerator
--- a/continuous_integration/environment-3.7.yml
+++ b/continuous_integration/environment-3.7.yml
@@ -13,7 +13,7 @@ dependencies:
   - pytest-flake8==1.0.4
   - dask==2.8.1
   - numpy==1.15.4
-  - scipy==1.2.0
+  - scipy==1.4.0
   - scikit-image==0.14.2
   - pims==0.4.1
   - slicerator
--- a/continuous_integration/environment-3.8-dev.yml
+++ b/continuous_integration/environment-3.8-dev.yml
@@ -13,7 +13,7 @@ dependencies:
   - pytest-flake8==1.0.4
   - dask==2.8.1
   - numpy==1.17.3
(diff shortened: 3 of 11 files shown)


GenevieveBuckley:
Hi @anlavandier 

> There is also a docstring for the function though I don't understand why dask and dask_image build different HTML pages with the exact same docstring.

I'm not entirely sure what you mean here. If it's a visual difference you're talking about, maybe that's due to [which sphinx theme is being used](https://github.com/dask/dask-image/pull/245). If your question is more about why the `dask-image` docs exist at the subdomain https;//image.dask.org then that is because this is a separate project. (John started `dask-image`, and then after it became relatively feature complete and stable it was shifted over to live under the Dask umbrella. That's typically how new Dask projects are born. Because everything was separate right from the very beginning, that's why some of the website or infrastructure stuff is different).

> Finally, while I was directed here by the core dask team, I personally think that this is even more confusing than adding the convolution in `dask.array` because it has basically nothing to do with `scipy.ndimage` or image processing in general.

Yes, I think I agree with you here. The scope of `dask.array` is intended to match numpy array functions. The scope of `dask-image` is intended to match `scipy.ndimage`. There isn't an existing project that is a clear "right place" to direct you towards. Possibly you could host/create a `dask-signal`?

GenevieveBuckley:
@jakirkham do you have thoughts?

jbigot:
Hello,

@anlavandier is implementing this as part of his internship at "Maison de la Simulation" a High-performance computing laboratory in France.

We would be glad to host a dedicated `dask-signal` repository. However we don't plan on putting much resources in porting more functions of `scipy.signal` to Dask. We just happen to use convolutions that are scattered all over numpy and scipy.

Do you think it makes sense to create such a repo? On the plus side, it would offer a location to gather potential `scipy.signal` porting contributions. On the other hand, it would be rather empty at least at first and unless many contributions appear.

Another option would be to expand `dask-image` into something like `dask-scipy`. Any thoughts on your side?

GenevieveBuckley:
Starting off under your own namespace sounds best. I'm glad you're thinking about how to share with the community @jbigot

PR #245, opened 2021-10-05
Use new Dask docs theme

jacobtomlinson:
Bumping the minimum docs theme to use the new version. This isn't strictly necessary but I wanted to raise a PR to trigger the rtd preview and check everything looks ok.

xref dask/dask-sphinx-theme#54

### DIFF
--- a/environment_doc.yml
+++ b/environment_doc.yml
@@ -7,10 +7,10 @@ dependencies:
   - pip==20.2.4
   - wheel==0.35.1
   - sphinx==3.2.1
-  - dask-sphinx-theme==1.3.5
+  - dask-sphinx-theme>=2
   - dask==2.8.1
   - numpy==1.15.4
   - pims==0.4.1
   - slicerator==0.9.8
   - pip:
-    - slicerator==0.9.8
+      - slicerator==0.9.8


jacobtomlinson:
I noticed that PR builds weren't enabled for this repo. I've turned that on too.

environment_doc.yml
```diff
@@ -7,10 +7,10 @@ dependencies:
...
   - sphinx==3.2.1
-  - dask-sphinx-theme==1.3.5
   - dask==2.8.1
   - numpy==1.15.4
   - pims==0.4.1
   - slicerator==0.9.8
   - pip:
-    - slicerator==0.9.8
+      - slicerator==0.9.8
+      - dask-sphinx-theme>=2
```
jakirkham, line 16: Would it make sense to still use the Conda package?

jakirkham (commented):
Thanks Jacob! 😄 Had one question below

environment_doc.yml
jacobtomlinson: This is the only project I've seen that uses the conda package. Everywhere else seems to use pip. So this change was partly for consistency, and partly because conda doesn't seem to have v2 yet for some reason.

PR #247, opened 2021-11-19
Switch imread function to dask.array.image method

GenevieveBuckley:
Work in progress PR for discussion

Benchmarking has shown that the `dask.array.image.imread` function is faster than the implementation here. We'd like to switch to using that one instead, the main question here is how disruptive or not it will be for our users.


jakirkham:
Sorry I'm probably missing something, but what benchmarking was done?

PR #248, opened 2021-12-15
Set up gpuCI testing on PRs

charlesbluca:
This PR adds the required build files to run gpuCI on all PRs opened in this repo; based on these files, the `cupy` marked tests in this repo would be run in an environment with:

- ubuntu 18.04
- python 3.8
- cudatoolkit 11.2
- the latest pre-release of cupy
- (note that since tests here do not depend on a RAPIDS release, `RAPIDS_VER` has no impact on the environment)

Some additional things that should be done:

- [ ] adding documentation for gpuCI
- [ ] adding a description for the `cupy` label being used by pytest in gpuCI

Note that gpuCI is not currently enabled on this repo - ideally, that would be done after merging these files in. To see an example of the gpuCI checks running, view https://github.com/charlesbluca/dask-image/pull/1.

cc @jakirkham @GenevieveBuckley

### DIFF
--- a/continuous_integration/gpuci/axis.yaml
+++ b/continuous_integration/gpuci/axis.yaml
@@ -1,0 +1,13 @@
+PYTHON_VER:
+- "3.8"
+
+CUDA_VER:
+- "11.2"
+
+LINUX_VER:
+- ubuntu18.04
+
+RAPIDS_VER:
+- "22.02"
+
+excludes:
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
Thank you @charlesbluca, this is very much appreciated!

The setup files look good to me, so I'll merge them and then we can move on to the next stage of testing.

PR #188, opened 2021-02-08
GitHub actions testing

GenevieveBuckley:
Switching continuous integration to GitHub Actions

Closes https://github.com/dask/dask-image/issues/178

PR #257, opened 2022-01-31
Remove/add testing for python 3.6/3.9, update CI pinnings

charlesbluca:
While working on [bumping gpuCI to python 3.9 for the Dask repos](https://github.com/dask/dask/pull/8642/files), I noticed that this repo only tests 3.6-3.8; given that dask/distributed currently test 3.7-3.9 and are currently planning to move to 3.8-3.10, I figured it might be worthwhile to explore adding 3.9 (and later 3.10) testing to this repo.


charlesbluca:
rerun tests

continuous_integration/environment-3.9.yml
```diff
@@ -0,0 +1,19 @@
...
+  - coverage>=4.5.4
+  - flake8>=3.7.9
+  - pytest>=5.3.1
+  - pytest-flake8>=1.0.4
+  - dask>=2.8.1
+  - numpy>=1.17.3
+  - scipy>=1.3.2
+  - scikit-image>=0.16.2
+  - pims>=0.4.1
+  - slicerator
```
charlesbluca, line 19: I've unpinned dependencies for 3.9, as the explicit pinning was causing issues with the environment solve; is there any reason to keep the pinnings on the 3.7 / 3.8 environments?

GenevieveBuckley: Just to double check, you've removed the pinned version from slicerator here, and that's the only change? I'm ok with that.

I would prefer to keep the pinnings on the older 3.8/3.8 environments, on the basis of "if it ain't broke, don't fix it". That's kind of a conservative stance, but we haven't received any reports of this causing issues for people so I'm not inclined to fiddle with that.

GenevieveBuckley: We should give @jakirkham a day or two to voice an objection if he has one (I don't *think* that's especially likely, but I certainly don't want to speak for him here, and don't have all the background context he does)

GenevieveBuckley:
Thank you @charlesbluca, I appreciate you taking this on

continuous_integration/environment-3.9.yml
GenevieveBuckley: Thanks for clarifying! In that case, I think we *do* want to pin the dependencies for 3.9 testing.

You're right that allowing the test CI to use the latest versions of dask, numpy and scipy might give us an earlier indication of compatibility problems, but to be honest we just don't have the bandwidth over at this project to be proactive about that. In addition, the CI is typically run quite infrequently here (like months apart) and there can be a *lot* of changes with other library updates in the interim, so when any failures due to package updates do happen it will be very difficult to narrow down where the problem is coming from. Those constraints are the reason for the decision here.

jakirkham: Agree with Genevieve. The initial idea of pinnings here was to minimize surprising CI failures due to dependency updates.

Understand why pinnings for Python 3.8 wouldn't copy over to Python 3.9. That said, it would be good to find a set of frozen dependencies that do make sense for Python 3.9 and include them here.

Also would be good to keep a set of frozen dependencies for other Python versions tested. Optionally we may take a moment here to refresh them.

charlesbluca: Thanks for the detailed explanation @GenevieveBuckley - in that case, I will go ahead and pin the 3.9 environment to the versions picked up in the environment creation.

> Optionally we may take a moment here to refresh them.

Happy to do that here if we're okay with keeping that all in one PR 🙂 also while we're refreshing the pinnings should I go ahead and pin `slicerator` since that was unpinned up until now?

charlesbluca:
rerun tests

continuous_integration/environment-3.9.yml
jakirkham: > should I go ahead and pin `slicerator` since that was unpinned up until now?

I don't recall the full history here, but we did use to have this pinned and that changed in PR ( https://github.com/dask/dask-image/pull/188 ). Perhaps there was some issue with pinning previously?

Would think we would want that pinned, but will defer to @GenevieveBuckley on this as she may know more 🙂

GenevieveBuckley: The reason slicerator is unpinned appears to be that when I was trying to make the github actions CI run, the pinning was causing me problems. There is a commit here from me called ["Yep, we really do need slicerator unpinned "](https://github.com/dask/dask-image/pull/188/commits/47b242ae607ec0a905451d2e4d180e7832e3b419), so you can see I tried slicerator both pinned and unpinned in PR https://github.com/dask/dask-image/pull/188

Mind you, I just unpinned it entirely, instead of trying to work out what other pinned version would work. I cut corners to finish the job quicker.

GenevieveBuckley (approved):
Looks good to me!

GenevieveBuckley:
Speaking of dependencies, should we bump the minimum allowable dask version? I just got a dependabot alert in another repo indicating an update to dask=2021.10.0 is necessary:

> An issue was discovered in Dask (aka python-dask) through 2021.09.1. Single machine Dask clusters started with dask.distributed.LocalCluster or dask.distributed.Client (which defaults to using LocalCluster) would mistakenly configure their respective Dask workers to listen on external interfaces (typically with a randomly selected high port) rather than only on localhost. A Dask cluster created using this method (when running on a machine that has an applicable port exposed) could be used by a sophisticated attacker to achieve remote code execution.

Bumping the dask version would need to happen in the setup.py file here https://github.com/dask/dask-image/blob/8e784e044bc268e6574ebc2e55529c64f966a589/setup.py#L32

Figured I should mention it on this thread, since that might throw a lot of @charlesbluca carefully crafted pinned versions for the CI out of whack too.

continuous_integration/environment-3.9.yml
charlesbluca: Thanks for the context! Things seem to be working now with it pinned to 1.0.0, so we should be safe to proceed with it pinned from now on

charlesbluca:
Thanks for the heads up! I think this PR should be good as is if you bump the dask dependency, as I pinned to `2022.1.1` (the most recent compatible version).

We should probably opt to bump Dask either as a follow up to or directly in this PR, since the change would break the current CI environments that are pinned to an older version of Dask:

https://github.com/dask/dask-image/blob/8e784e044bc268e6574ebc2e55529c64f966a589/continuous_integration/environment-3.8.yml#L14

jakirkham:
If we are bumping Dask, there is old code we should cleanup. Do we want to do this here? If not, would it make sense to keep an older Dask version in at least some CI tests?

charlesbluca:
I'm okay with pinning the 3.7 and/or 3.8 environment(s) to an older version of Dask and updating the other dependencies accordingly - did you have a preference on a specific Dask version? If not, we can always go back to 2.8.1

jakirkham:
The original pin seems like a reasonable choice

GenevieveBuckley:
> If we are bumping Dask, there is old code we should cleanup. Do we want to do this here? If not, would it make sense to keep an older Dask version in at least some CI tests?

I think code cleanup is a lower priority than security. So we should prioritize bumping the version of Dask (either in this PR or a new one, whatever is quickest/easiest), and then if necessary we can defer a code cleanup PR to a later time if need be.

jakirkham:
The newest Dask version is used in the Python 3.9 job. So this is getting tested (users of course can install the newer version as well).

So this is a question of continuing to make things work for the versions we claim support for.

Again no objection to dropping old code and bumping the version, but maybe that should be deferred to a separate PR with those changes?

GenevieveBuckley:
Thank you @charlesbluca 
This is excellent to have. I especially like the switch to using mamba, that looks like it will reduce the CI run times quite a bit.

Issue #229, opened 2021-05-17
One `imread` to rule them all


GenevieveBuckley:
A lot of people have put a lot of effort into `imread` lately. This is great, and it's really helped. However, we've still got a way to go.

This is where I see the four major areas problems pop up in:

1. Read image data into Dask arrays accurately. We need more simple test cases here. Bug report: https://github.com/dask/dask-image/issues/220

2. Reduce confusion. Currently, there are multiple implementations of a dask `imread` function. The two most easily confused are `dask_image.imread.imread()` and `dask.array.image.imread()`. We need to figure out which is best, and only use that one.

3. Read data in fast. For that, we'll need to have some proper benchmarks, and run them routinely as part of the CI. This will help us decide (2) above. Previous discussion:
    * Imread performance issue https://github.com/dask/dask-image/issues/181
    * Getting movie files into Dask efficiently https://github.com/dask/dask-image/issues/134

4. Process the image data fast, too. For that to happen, we need smart default choices for how we chunk image data in dask arrays. Jackson Maxfield Brown describes the problem well in [this short video here](https://www.youtube.com/watch?v=LNa_gGpSnvc&list=PLJ0vO2F_f6OBAY6hjRHM_mIQ9yh32mWr0&index=8)

GenevieveBuckley:
The first step is getting a benchmark script going.

We need to:

- [ ] Report benchmark results for `dask_image.imread.imread()` and `dask.array.image.imread()` (for an apples to apples comparison, you might need to explictly pass `pims.open` as a keyword argument to `dask.array.image.imread()`)
- [ ] Add benchmarking to run on our CI

JacksonMaxfield:
> 3. Read data in fast. For that, we'll need to have some proper benchmarks, and run them routinely as part of the CI. This will help us decide (2) above. Previous discussion:
> 
>     * Imread performance issue #181
> 
>     * Getting movie files into Dask efficiently #134

Highly recommend using [asv](https://asv.readthedocs.io/en/stable/index.html). We use it on `aicsimageio` to get our pure reading benchmarks and after working on the Dask Summit presentation I also just added the example from my slides as a benchmark suite and a "LibCompareSuite" to monitor aicsimageio and dask-image performance.

* [our benchmark code for general IO suites](https://github.com/AllenCellModeling/aicsimageio/blob/main/benchmarks/benchmark_image_containers.py)
[...]

PR #265, opened 2022-07-21
Use natural sorting in  `imread(...)` when globbing multiple files 

VolkerH:
`dask_image.imread.imread` can include glob strings in the filename argument.

Currently, the results of the glob operation are sorted using `sorted`, a behviour that is not described in the doctring.

`sorted` sorts alphabetically which causes files such as `10.tif`, `9.tif` to be read in this order, which is counter-intuitive.
A common strategy adopted by users to circumenvent such file name problems is to zero-pad all numbers in a string. 

This PR adds a `sortfunc` argument to to `imread`, which allows to pass in a custom sort function. A use case for this  is to pass in a function that performs natural sort, which will sort strings of numbers numerically. Such a use-case is shown in the included test.

The docstring for imread has been updated to include the new argument, which also remedies the current undocumented behaviour.
Files have been checked with `flake8`.

### DIFF
--- a/dask_image/imread/__init__.py
+++ b/dask_image/imread/__init__.py
@@ -25,8 +25,8 @@ def imread(fname, nframes=1, *, arraytype="numpy", sortfunc=sorted):
         Number of the frames to include in each chunk (default: 1).
     arraytype : str, optional
         Array type for dask chunks. Available options: "numpy", "cupy".
-    sortfunc: Callable
-        A function for sorting the glob results.
+    sortfunc: Callable, optional
+        A function for sorting the glob results, default is "sorted".
 
     Returns
     -------
--- a/dask_image/imread/__init__.py
+++ b/dask_image/imread/__init__.py
@@ -10,7 +10,7 @@ import pims
 from . import _utils
 
 
-def imread(fname, nframes=1, *, arraytype="numpy"):
+def imread(fname, nframes=1, *, arraytype="numpy", sortfunc=sorted):
     """
     Read image data into a Dask Array.
 
@@ -25,6 +25,8 @@ def imread(fname, nframes=1, *, arraytype="numpy"):
         Number of the frames to include in each chunk (default: 1).
     arraytype : str, optional
         Array type for dask chunks. Available options: "numpy", "cupy".
+    sortfunc: Callable
+        A function for sorting the glob results.
 
     Returns
     -------
@@ -65,7 +67,7 @@ def imread(fname, nframes=1, *, arraytype="numpy"):
         )
 
     # place source filenames into dask array
-    filenames = sorted(glob.glob(sfname))  # pims also does this
+    filenames = sortfunc(glob.glob(sfname))  # pims also does this
     if len(filenames) > 1:
         ar = da.from_array(filenames, chunks=(nframes,))
         multiple_files = True
@@ -83,7 +85,6 @@ def imread(fname, nframes=1, *, arraytype="numpy"):
         arrayfunc=arrayfunc,
         meta=arrayfunc([]).astype(dtype),  # meta overwrites `dtype` argument
     )
-
     return a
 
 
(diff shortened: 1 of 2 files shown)


GPUtester:

GenevieveBuckley:
add to allowlist

GenevieveBuckley:
Do you anticipate needing to use a sort function besides natural sort?

I'd be happy to straight up replace `sorted` with [natsort](https://github.com/SethMMorton/natsort#natsort). It's an extra dependency, but I think the benefit would be more than worth it.

I'm less keen to change the API to allow users to pass in arbitrary sorting functions, although I see why you suggest it. Instead, I think it would be cleaner to allow the function to accept a list of strings or Paths. Then the user could modify this list in any arbitrary way before providing it to dask-image, and it's much easier for the user to inspect. 

Ideally, we'd do both these things. Then there would be a sensible default with glob and natsort, plus a more flexible alternative if necessary.

GenevieveBuckley:
> I'd be happy to straight up replace sorted with [natsort](https://github.com/SethMMorton/natsort#natsort). It's an extra dependency, but I think the benefit would be more than worth it.

Oh, even easier! I didn't realize `tifffile` had a `natural_sorted` function (`from tifffile import natural_sorted`). We already require tifffile as a dependency, so that's even easier to switch to using. (Thanks for teaching me something new today 😄  )

VolkerH:
I agree that natural sort would be the most sensible default. 
I just did not want to change the existing behaviour in case any legacy code is relying on alphabetical sort (unlikely, most of those legacy cases would probably be bugs). 
Passing in a list of files would be another option (I  also thought of that).

Yes, `tifffile` has a lot of hidden gems. Most of the documentation is in the actual `tifffile` python source code, so it is a bit hidden.

GenevieveBuckley:
How about you edit this PR to remove the API change, and instead do a straight replacement  of the current behaviour with `tifffile.natural_sort`. I'd be happy to accept that more or less as-is. 

The current behaviour is clearly a bug, and it is very unlikely anybody is relying on it for legacy code. 

I'd also be happy to see a PR allowing users to pass in a list of filenames, but that should be a separate PR. (You're welcome to try making that too if you like, but I also don't expect you to. Your problem will likely be solved after this first PR and you'll want to get back to what you were doing earlier)

> Yes, tifffile has a lot of hidden gems. Most of the documentation is in the actual tifffile python source code, so it is a bit hidden.

I have seen that documentation in the source code, perhaps it's time to go back and see what other cool stuff I missed!

VolkerH:
I made natural sorting the default and removed the ability to pass in a custom sort function as expected.

As side notes (probably better placed in two separate issues):
* The documentation should explain what the difference between `dask_image.imread.imread` and `dask.array.image.imread` is. I know that there muse a difference because ...
* .... I noticed that `dask_image.imread.imread` fails for certain `.tif` files where `dask.array.image.imread` works just fine. The source of the error isn't in `dask_image` but in `pims` and has to do with a datetime string formatting in the `tif` metadata that is not expected by `pims`. I will try and make a self-contained example and raise an issue in `pims`.

GenevieveBuckley (approved):
Thank you @VolkerH, this is a useful improvement

GenevieveBuckley:
> * The documentation should explain what the difference between `dask_image.imread.imread` and `dask.array.image.imread` is and when to use which. I know that there must a difference because ...

Yes, this is perhaps better discussed in https://github.com/dask/dask-image/issues/229
I agree, we should have one reliable way to do it. The current situation is confusing.

Minor differences include:
* `dask.array.image.imread` will make one chunk per filename, which is not always a great choice (eg: you might have a single massive file with multiple timepoints in it, etc.)
* `dask.array.image.imread` uses the scikit-image imread by default (although you can pass in whatever reader you like with a keyword argument, which is an excellent feature). Whereas dask-image uses pims, and you don't have a choice in the matter. 
    *For the future, I think imageio now allows you to read file metadata without reading the *entire* file, so it would be good to make dask-image depend on that instead of pims (partly also because that's a larger and more active project)

PR #267, opened 2022-08-07
[Refactoring] making the code more Pythonic

maldil:
The goal of this PR is to make the code more Pythonic by utilizing Python with statements.

### DIFF
--- a/dask_image/_version.py
+++ b/dask_image/_version.py
@@ -138,21 +138,20 @@ def git_get_keywords(versionfile_abs):
     # _version.py.
     keywords = {}
     try:
-        f = open(versionfile_abs, "r")
-        for line in f.readlines():
-            if line.strip().startswith("git_refnames ="):
-                mo = re.search(r'=\s*"(.*)"', line)
-                if mo:
-                    keywords["refnames"] = mo.group(1)
-            if line.strip().startswith("git_full ="):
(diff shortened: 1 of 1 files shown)


GPUtester:

GenevieveBuckley:
Hi @maldil, thanks for dropping by!
 
FYI, this file was automatically generated by [versioneer](https://github.com/python-versioneer/python-versioneer), so typically I wouldn't edit that.

https://github.com/dask/dask-image/blob/main/versioneer.py#L427-L428

If you think it's useful, you could suggest your contribution over at the main versioneer project: https://github.com/python-versioneer/python-versioneer
That way, any improvements can be used by everyone, instead of only in a single project.

jakirkham:
It is also possible a new version of versioneer has solved this, in which case we can [upgrade]( https://github.com/python-versioneer/python-versioneer#updating-versioneer )

GenevieveBuckley:
True, I think the versioneer version here is probably from when this repository was created.

GenevieveBuckley:
We have versioneer version 0.18 (released September 2016) and the current version is 0.22 (released March 2022). According to the [docs](https://github.com/python-versioneer/python-versioneer/blob/master/UPGRADING.md) there is "nothing special" we need to do, other than [follow the instructions here](https://github.com/python-versioneer/python-versioneer#updating-versioneer)

Issue #269, opened 2022-09-15
New release


thewtex:
Hi @GenevieveBuckley @jakirkham !

Checking into the scipy API warnings described at https://github.com/spatial-image/multiscale-spatial-image/issues/58, they are resolved with `dask-image` `main`. Is it possible to cut a new release? :pray:

GenevieveBuckley:
Release in progress now :)

Link to the fix for the scipy deprecations:
* Issue: https://github.com/dask/dask-image/issues/260
* PR fix: https://github.com/dask/dask-image/pull/261

GenevieveBuckley:
Here we go: https://github.com/dask/dask-image/releases/tag/v2022.09.0

Try it out and tell me if there are any problems.

The release notes say "This will automatically upload the release to PyPI, and will also automatically begin the process to release the new version on conda-forge."
I'll have to check https://github.com/conda-forge/dask-image-feedstock for the bot's PR tomorrow.

GenevieveBuckley:
The new release should be available on conda-forge now I've merged this PR: https://github.com/conda-forge/dask-image-feedstock/pull/11

thewtex:
@GenevieveBuckley thank you, thank you!! Much greener now! 🙏 💚

PR #270, opened 2022-09-19
Release notes for v2022.09.0

GenevieveBuckley:
Release notes for v2022.09.0

Release requested by Matt McCormick here: https://github.com/dask/dask-image/issues/269

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,46 @@
 History
 =======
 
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
All CI checks green (except the readthedocs build, but I'm not worried about waiting for that one since we already show the docs from the master branch at https://image.dask.org/en/latest/ and there's nothing in this PR that should negatively affect it).

Issue #277, opened 2023-01-10
Readthedocs build is failing


GenevieveBuckley:
## What happened
I've noticed the Readthedocs build is failing on the CI checks for PR 237.

* Here's a link to a recent failing build: ttps://readthedocs.org/projects/dask-image/builds/19108499/
* The error message says "We encountered a problem with a command while building your project. To resolve this error, double check your project configuration and installed dependencies are correct and have not changed recently."
* And the python traceback error "ImportError: cannot import name 'environmentfilter' from 'jinja2' (/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/site-packages/jinja2/__init__.py)"

<details>
<summary>Expand for build failure details:</summary>

```

ttps://readthedocs.org/projects/dask-image/builds/19108499/
Build #19108499
Pull Request #237 (431de07f34d2391c957bd01a94df9863f1bb7699) 

Build failed

Error
We encountered a problem with a command while building your project. To resolve this error, double check your project configuration and installed dependencies are correct and have not changed recently.

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/site-packages/sphinx/__main__.py", line 13, in <module>
    from sphinx.cmd.build import main
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/site-packages/sphinx/cmd/build.py", line 25, in <module>
    from sphinx.application import Sphinx
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/site-packages/sphinx/application.py", line 42, in <module>
    from sphinx.registry import SphinxComponentRegistry
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-image/conda/237/lib/python3.7/site-packages/sphinx/registry.py", line 24, in <module>
    from sphinx.builders import Builder
[...]

PR #278, opened 2023-01-10
Pin to jinja2<3.1 to avoid Readthedocs build error

GenevieveBuckley:
Closes issue https://github.com/dask/dask-image/issues/277

### DIFF
--- a/continuous_integration/environment-doc.yml
+++ b/continuous_integration/environment-doc.yml
@@ -8,7 +8,7 @@ dependencies:
   - pip==22.3
   - wheel==0.37.1
   - sphinx==5.3.0
-  - jinja2==3.1.2
+  - jinja2<3.1
   - dask==2.8.1
   - numpy==1.23.4
   - scipy==1.9.2


GenevieveBuckley:
Note: I didn't test this locally before opening this PR. I'm hoping the CI checks will run the Readthedocs build with the updated yaml file and test it for us.

Ordinarily I'd test any changes to yaml environment files locally, just to check the dependencies can resolve. But I wasn't able to do that this time, because on my M1 Mac conda can't create even the current environment-doc.yml
```
Encountered problems while solving:
  - nothing provides msgpack-python <1.0.0 needed by distributed-2.10.0-py_0
```
I think that's just an M1 Mac problem, so it probably won't matter since the CI uses linux. 

@jakirkham you did a lot of work on https://github.com/dask/dask-image/pull/273 so you're the most expert on the documentation CI setup right now - please let me know if you think I should be doing anything differently

jakirkham:
Sorry for the slow reply Genevieve 😞

Thanks for digging into this issue! 🙏

Remember some weird things going on with Sphinx, but have forgotten what they were now. Am wondering if maybe this was a dependency issue that got cleared up somewhere? Unfortunately search is failing me atm

GenevieveBuckley:
All the CI checks are passing, so I'm going to merge this. Fingers crossed it resolves the problems.

Issue #232, opened 2021-05-21
Increase minimum Dask version and remove try/except imports


GenevieveBuckley:
We have this try/except statement for imports here:

https://github.com/dask/dask-image/blob/bbe73c64ec16781053858c0408e9ca6e4af6fb9d/dask_image/ndmeasure/_utils/__init__.py#L14-L17

According to John:

> Yeah this was renamed in Dask 1.1.0 ( https://github.com/dask/dask/pull/4348 ). If we are comfortable making that a minimum requirement, we can drop that and just use `dask.array.blockwise` below

_Originally posted by @jakirkham in https://github.com/dask/dask-image/pull/227#discussion_r636444657_

Reply from Gen:
> I suspect that changing the minimum requirement for Dask is going to involve some fiddling with the environment files we use for the CI, so I'll open an issue & that can be a separate PR.

GenevieveBuckley:
Thank you @jakirkham

Issue #258, opened 2022-02-12
Bump the minimum allowable version of Dask


GenevieveBuckley:
Should we bump the minimum allowable dask version? I just got a dependabot alert in another repo indicating an update to dask=2021.10.0 is necessary:

> An issue was discovered in Dask (aka python-dask) through 2021.09.1. Single machine Dask clusters started with dask.distributed.LocalCluster or dask.distributed.Client (which defaults to using LocalCluster) would mistakenly configure their respective Dask workers to listen on external interfaces (typically with a randomly selected high port) rather than only on localhost. A Dask cluster created using this method (when running on a machine that has an applicable port exposed) could be used by a sophisticated attacker to achieve remote code execution.

Bumping the dask version would need to happen in the setup.py file here https://github.com/dask/dask-image/blob/8e784e044bc268e6574ebc2e55529c64f966a589/setup.py#L32

_Originally posted by @GenevieveBuckley in https://github.com/dask/dask-image/issues/257#issuecomment-1033583468_

GenevieveBuckley:
Here's a copy of the rest of the comments on this topic from the [original thread]()

[charlesbluca replies...](https://github.com/dask/dask-image/pull/257#issuecomment-1034061010)
> Thanks for the heads up! I think this PR should be good as is if you bump the dask dependency, as I pinned to `2022.1.1` (the most recent compatible version).
>
> We should probably opt to bump Dask either as a follow up to or directly in this PR, since the change would break the current CI environments that are pinned to an older version of Dask:
>
> https://github.com/dask/dask-image/blob/8e784e044bc268e6574ebc2e55529c64f966a589/continuous_integration/environment-3.8.yml#L14

[jakirkham says...](https://github.com/dask/dask-image/pull/257#issuecomment-1034105422)
>If we are bumping Dask, there is old code we should cleanup. Do we want to do this here? If not, would it make sense to keep an older Dask version in at least some CI tests?

[GenevieveBuckley says...](https://github.com/dask/dask-image/pull/257#issuecomment-1035748954)
> I think code cleanup is a lower priority than security. So we should prioritize bumping the version of Dask (either in this PR or a new one, whatever is quickest/easiest), and then if necessary we can defer a code cleanup PR to a later time if need be.

[jakirkham says...](https://github.com/dask/dask-image/pull/257#issuecomment-1035891205)
> The newest Dask version is used in the Python 3.9 job. So this is getting tested (users of course can install the newer version as well where most security issues are likely to surface).
>
[...]

PR #286, opened 2023-03-13
Require Dask 1.1.0+

jakirkham:
Fixes https://github.com/dask/dask-image/issues/232

Bumps the Dask minimum version to 1.1.0, which in turn allows some special casing logic to be dropped.

### DIFF
--- a/dask_image/ndmeasure/_utils/__init__.py
+++ b/dask_image/ndmeasure/_utils/__init__.py
@@ -5,11 +5,6 @@ import dask
 import dask.array as da
 import numpy as np
 
-try:
-    from dask.array import blockwise as da_blockwise
-except ImportError:
-    from dask.array import atop as da_blockwise
-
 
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
This is great, thank you @jakirkham 

Since Dask [version 1.1.0 was released on the 19th of January 2019](https://github.com/dask/dask/releases/tag/1.1.0) (over four years ago now), this requirement is very unlikely to negatively affect anyone.

jakirkham:
Happy to help!

We could also bump this more if we like

This just seemed like the simplest change that already had some consensus around it

GenevieveBuckley:
> We could also bump this more if we like

I forgot about the security update suggested by dependabot here: https://github.com/dask/dask-image/issues/258 (oops!) We should bump it again to Dask>=2021.10.0

I'll make a PR. This one went smoothly, so hopefully the next one will go smoothly too.

Issue #271, opened 2022-10-19
Publish .whl to PyPI


thewtex:
Hi @GenevieveBuckley @jakirkham ,

Another humble request :pray: -- We are trying to use dask-image in Pyodide / JupyterLite and its needs the `*.whl` file. Could this please be published to PyPI? It looks like there is only the sdist now.  It can be generated with, for example `cd ~/src/dask-image; pip wheel .`

Thank you!

CC: @bnmajor

jakirkham:
Think this would just be a matter of updating this line. Maybe an easy PR for someone? 😉

https://github.com/dask/dask-image/blob/b8ff7d9d370750a0074cd7d973329f79669fbc05/.github/workflows/python-publish.yml#L30

thewtex:
> Think this would just be a matter of updating this line.

:eyes: 

> Maybe an easy PR for someone? 

:raising_hand: #272 

:-)

jakirkham:
Thanks Matt! 🙏

Now that is in we should be good to do this as part of the next release.

thewtex:
Thank you, John!

jakirkham:
@GenevieveBuckley, do you recall what we need to do to create a new release?

GenevieveBuckley:
The release guide is here: https://github.com/dask/dask-image/blob/main/docs/release/release_guide.rst

If you notice anything that needs to be updated in that doc, feel free to edit it. (I think the only thing we don't really do is make release candidates, mostly because we don't really have people available to test them, so no need to make the process longer than it has to be)

GenevieveBuckley:
[Release 2023.03.0](https://github.com/dask/dask-image/releases/tag/v2023.03.0) of dask-image is now available, and the .whl files [are now uploaded to PyPI](https://pypi.org/project/dask-image/#history).

Let us know if you need anything else, Matt.

PR #284, opened 2023-03-13
Update supported python versions to 3.8, 3.9, 3.10, & 3.11 (drop python 3.7)

GenevieveBuckley:
The main dask/dask repo [supports python versions 3.8, 3.9, 3.10, & 3.11](https://github.com/dask/dask/blob/main/.github/workflows/tests.yml#L26-L27). The dask-image project should ideally match this, for consistency.

This PR:
* drops python 3.7 from the test CI
* adds python 3.10 & 3.11 to the test CI
* updates the pull request template

PR #288, opened 2023-03-14
Security update: Dask v2021.10.0 as minimum allowable version

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/258

Dependabot suggests bumping the minimum allowable version of dask to v2021.10.0, as a security update.

> An issue was discovered in Dask (aka python-dask) through 2021.09.1. Single machine Dask clusters started with dask.distributed.LocalCluster or dask.distributed.Client (which defaults to using LocalCluster) would mistakenly configure their respective Dask workers to listen on external interfaces (typically with a randomly selected high port) rather than only on localhost. A Dask cluster created using this method (when running on a machine that has an applicable port exposed) could be used by a sophisticated attacker to achieve remote code execution.

Once this is done, we should probably also publish another release to PyPI (which would also solve issue https://github.com/dask/dask-image/issues/271)

### DIFF
--- a/continuous_integration/environment-3.11.yml
+++ b/continuous_integration/environment-3.11.yml
@@ -1,20 +1,20 @@
-name: dask-image-testenv-py311
+name: dask-image-testenv
 
 channels:
   - conda-forge
 
 dependencies:
   - python=3.11.*
-  - pip
-  - wheel
-  - coverage
-  - flake8
-  - pytest
-  - pytest-cov
-  - pytest-flake8
-  - dask
-  - numpy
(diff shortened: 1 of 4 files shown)


GenevieveBuckley:
This PR also fixes some things from https://github.com/dask/dask-image/pull/284:
- somehow I left off the pinned version numbers in `continuous_integration/environment-3.11.yml`
- `continuous_integration/environment-latest.yml` should also have been updated to have the same contents as the environment-3.11.yml file

PR #290, opened 2023-03-14
Group all imread functions together in the same file

GenevieveBuckley:
This PR groups all imread functions into the same file.

I find it annoying to have all the imread functions together, *except* for one little two line function which is split onto another page. It also makes it weirdly difficult to fully understand what is happening with the code, so I'm changing it.

### DIFF
--- a/dask_image/imread/__init__.py
+++ b/dask_image/imread/__init__.py
@@ -8,8 +8,6 @@ import numpy as np
 import pims
 from tifffile import natural_sorted
 
-from . import _utils
-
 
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
All CI checks passing, merging this.

m-albert:
Completely agree!

jakirkham:
Yeah think this was here as I was concerned about having this function leaking into the public API (since it is intended to be private). Though perhaps prefixing the function name with an `_` is good enough

PR #291, opened 2023-03-14
Combine CI workflows for testing and release upload to PyPI

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/147

We want the tests to run and pass before a new release is uploaded to PyPI. This will be very useful, especially since activity on this repo is infrequent, so it's a good idea to re-run the whole test suite before releasing.

Combining the two CI workflows into the same file makes this easy, since now the PyPI upload is an optional extra step if the tests all pass and a tagged version has been created (either with `git --tag v2023.MM.DD` or by making a release from the github releases page).

We'll be able to see whether any of the tests fail and why in the [github actions tab](https://github.com/dask/dask-image/actions) for this repository. It makes it much easier to create a release candidate tag, check the tests pass and it's uploaded to PyPI, then make the official release.


GenevieveBuckley:
The new combined CI checks are passing, and we'll see how smoothly it performs at the next release (we should do that soon).

jakirkham:
Awesome! Thanks Genevieve 🙏

Agree it would be good to do a release soon

PR #294, opened 2023-03-14
Readd `environment-latest.yml` symlink

jakirkham:
Looks like GHA Windows builds are having some issues following this symlink (for [example]( https://github.com/dask/dask-image/actions/runs/4412687582/jobs/7732378054 )). Try removing and readding to see if that fixes the issue.


jakirkham:
Merging to fix up CI. Though if something else is needed, happy to follow up

GenevieveBuckley:
Thank you!

PR #295, opened 2023-03-14
Make tifffile regular requirement

m-albert:
I just realised that although `tifffile` is imported in `dask_image.imread.__init__`, it's only part of the test requirements. Indeed, this leads to an import error in a fresh environment.

This PR upgrades `tifffile >=2018.10.18` from a test requirement to a regular requirement in `setup.py`. Regarding the version, `2018.10.18`, which had been pinned as a test requirement, also provides `natural_sorted` as introduced in https://github.com/dask/dask-image/pull/265.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ requirements = [
     "numpy >=1.11.3",
     "scipy >=0.19.1",
     "pims >=0.4.1",
+    "tifffile >=2018.10.18",
 ]
 
 test_requirements = [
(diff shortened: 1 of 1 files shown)


GenevieveBuckley (approved):
I'm in favour of this change!

I'll suggest leaving it open for a day or so to let @jakirkham have a chance to respond, and if you don't hear any objections I'm happy for you to go ahead and merge it.

jakirkham:
Thanks Marvin for the PR and Genevieve for the review! 🙏

PR #299, opened 2023-03-27
Update release notes script

GenevieveBuckley:
The release notes script generated produced an error, so this small update makes it work again for newer versions of `pygithub`.

Honestly most of this functionality is now included as part of the Github release page (https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes) so we could potentially get rid of the script altogether.


GenevieveBuckley:
I'm not sure why the GPU CI seems unhappy today. However, this PR is unrelated to any of cupy/GPU code, so I'm going to ignore that for now.

Issue #308, opened 2023-03-30
pandas is required by find_objects function


GenevieveBuckley:
Pandas is a requirement of the `find_objects` function introduced in https://github.com/dask/dask-image/pull/240, but this wasn't actually added to the setup install requirement list. We should fix that.

PR #309, opened 2023-03-30
Add pandas requirement for find_objs function

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/308


setup.py
```diff
@@ -34,6 +34,7 @@ def run_tests(self):
     "scipy >=0.19.1",
     "pims >=0.4.1",
     "tifffile >=2018.10.18",
+    "pandas",
```
GenevieveBuckley, line 37: Should we give a specific minimum version? If so, how to choose one?

GenevieveBuckley:
I don't know why the GPU CI is encountering problems with python 3.8, but our github actions CI for python 3.8 is passing. 

@jakirkham is there any way I can see what package versions were installed into the GPU CI environment?

In any case, it seems like the quickest way forward is to go ahead and drop support for python 3.8 as discussed, and then move forward with this (PR here: https://github.com/dask/dask-image/pull/315)

jakirkham:
Looks like merging PR ( https://github.com/dask/dask-image/pull/315 ) introduced some conflicts (sorry about that)

Though it does sound like that PR will fix things here. Happy to revisit if not

setup.py
GenevieveBuckley: I'm choosing to set the minimum pandas version as 2.0.0. 
It's not a hard technical requirement, but it will make future maintenance easier for us later if we restrict it to pandas version 2 only.

GenevieveBuckley:
Hm, I had expected that thie GPU CI tests would pass now that I've merged the main branch, since https://github.com/dask/dask-image/pull/317 is passing.

jakirkham:
May need PR ( https://github.com/dask/dask-image/pull/318 ) as well, which was just merged. Updated this PR to pick up those changes

jakirkham (approved):

jakirkham:
Thanks Genevieve! 🙏

jakirkham:
rerun tests

(Python 3.10 build passed, but the Python 3.9 build ran into some error)

jakirkham:
Looks like that worked! 🎉

m-albert:
This might be a bit late.

I was wondering whether `pandas>=2.0.0` is required for the find_objects function? In case it's not, considering that pandas 2 was released relatively recently (and the problem we had with the new arrow strings) relaxing this might be useful to users.

GenevieveBuckley:
> I was wondering whether `pandas>=2.0.0` is required for the find_objects function? In case it's not, considering that pandas 2 was released relatively recently (and the problem we had with the new arrow strings) relaxing this might be useful to users.

I don't fully remember anymore, but I did think at one stage "oh, this is going to be a real pain to support both pandas 1 and 2, let's just make it 2". But I think that might have had more to do with getting our continuous integration environment files working well.

Because there's no particular evidence that anyone really needs this right now, and more and more people will be switching to pandas 2 over time, I don't think it's the right place for this project to be spending effort on at the moment.

m-albert:
You're completely right, there's no concrete evidence that anyone would need compatibility with `pandas<2`. While bringing this up I was clearly biased by the pyarrow problem (specific to pandas>=2), to which we found a different solution for now.

Thank you for commenting on this and I agree it's not worth spending effort here currently.

Issue #314, opened 2023-05-01
Drop python 3.8


GenevieveBuckley:
The Dask community is discussing dropping python 3.8 here https://github.com/dask/community/issues/315

I think we should follow that discussion and do the same thing here. It looks as though the consensus is leaning towards dropping python 3.8 soon-ish.

jakirkham:
SGTM

AIUI that would also fix CI. Is that right?

GenevieveBuckley:
I think so. 
I think it's only the python 3.8 Ci environment that has that particular problem with the pandas and dask versions being incompatible https://github.com/dask/dask-image/issues/312

jakirkham:
Think it would be ok if we wanted to drop Python 3.8 before dask does (since dask-image is downstream of dask)

PR #315, opened 2023-05-10
Drop Python 3.8, in accordance with NEP29 recommendation

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/314

This PR drops support for python 3.8

The [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) recommendation is to drop support for python 3.8 on April 14th 2023. We're about a month past that date now, so we're dropping support for it.

The second reason for doing this now is that our python 3.8 CI test environments are starting to have some problems with dependency version mismatches, and it's not a good use of maintainer time to figure out those issues when we're about to drop support for python 3.8 anyway.


GenevieveBuckley:
I don't understand why the GPU CI is returning build failures, when the github actions CI manages to create its environment just fine.

@jakirkham do you have any debugging tips? I'm not finding a lot of useful information in the Jenkins console output.

jakirkham:
cc @charlesbluca

continuous_integration/gpuci/axis.yaml
```diff
@@ -1,6 +1,6 @@
 PYTHON_VER:
-- "3.8"
-- "3.10"
+- "3.9"
+- "3.11"
```
charlesbluca, line 3: We're currently only publishing GPU CI images for python 3.9/3.10:

```suggestion
- "3.10"
```

Is there a specific need for images using python 3.11? If so can explore building them.

For reference, here are some relevant files for dask-image's GPU CI images:

- https://github.com/rapidsai/dask-build-environment/blob/main/ci/axis/dask.yaml - controls the matrix of images being built
- https://github.com/rapidsai/dask-build-environment/blob/main/dask_image/Dockerfile - controls the build steps of the images
- https://github.com/rapidsai/dask-build-environment/blob/main/dask_image/environment.yml - specifies the GPU-specific conda requirements to install in the CI environment

charlesbluca:
Did some digging into the python 3.9 failures on GPU and looks like those are also related to the images - since we are bumping the conda environment files that the images are built against in this PR, they don't yet reflect some of the changes that were made to unblock tests, specifically the Dask version bump.

I will note that locally I was able to unblock these failures by modifying the dask-image install command in the GPU test script:

https://github.com/dask/dask-image/blob/d571db7bba72a404aef18e8bb1d8d3c3d897bbbb/continuous_integration/gpuci/build.sh#L46-L47

As a `pip install` pulls the required Dask version.

continuous_integration/gpuci/axis.yaml
jakirkham: Went ahead and committed this change. We can revisit when there are Python 3.11 images

jakirkham:
After talking with Charles about this offline, it sounds like this CI issue should go away after these changes are in `main` (though not before). So going to go ahead and merge. Can create a trivial change PR afterwards to make sure things are working (or identify any remaining issues to fix)

jakirkham (approved):

jakirkham:
Thanks Genevieve and Charles! 🙏

jakirkham:
Submitted PR ( https://github.com/dask/dask-image/pull/317 ) to retest gpuCI now that this is merged

jakirkham:
Fixed one more Python version constrain in PR ( https://github.com/dask/dask-image/pull/336 )

PR #317, opened 2023-05-10
[NO MRG] Retest gpuCI

jakirkham:
Opening this PR to test gpuCI now that PR ( https://github.com/dask/dask-image/pull/315 ) has been merged

cc @GenevieveBuckley @charlesbluca

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+
 import sys
 
 import setuptools


charlesbluca:
rerun tests

jakirkham:
Thanks Charles! 🙏

Is there typically a lag between when a CI fix is merged and when other PRs can get the fix?

charlesbluca:
Yeah, currently the images are made on a nightly basis, so any changes to the environment files would only be reflected once the next scheduled run is complete; in this case, I manually triggered a rebuild of the images so that we could more immediately test the change.

Perhaps a good compromise between minimizing runtime and reflecting changes to the environment files would be to run `mamba update -f $ENV` at the beginning of each run?

jakirkham:
Ah ok. Thanks for the context (and the assist)! 🙏

That seems reasonable. Would that go after this step or somewhere else?

https://github.com/dask/dask-image/blob/c62a503fce87e602c829d32d6c90599c7ff0cc33/continuous_integration/gpuci/build.sh#L36-L38

GenevieveBuckley:
Thank you @charlesbluca and @jakirkham for tying to figure this out!

charlesbluca:
> Would that go after this step or somewhere else?

I don't think it should matter, but I suppose updating before activating the environment wouldn't hurt - I can open up a PR trying this out to see if there's any immediately obvious issues we would run into

jakirkham:
Toggling to retest with PR ( https://github.com/dask/dask-image/pull/318 ) merged

jakirkham:
Seems like things are working now as PR ( https://github.com/dask/dask-image/pull/309 ) also passed

Closing this out

Thanks again Charles! 🙏

PR #318, opened 2023-05-15
Update GPU conda environment before running tests

charlesbluca:
Attempt to update the conda environment in the GPU CI images before running tests, which should hopefully allow us to capture environment changes like those made in #315 without needing to rebuild images.

### DIFF
--- a/continuous_integration/gpuci/build.sh
+++ b/continuous_integration/gpuci/build.sh
@@ -33,15 +33,15 @@ env
 gpuci_logger "Check GPU usage"
 nvidia-smi
 
-gpuci_logger "Activate conda env"
+gpuci_logger "Update conda environment"
 . /opt/conda/etc/profile.d/conda.sh
+gpuci_mamba_retry update -n dask_image -f "$WORKSPACE/continuous_integration/environment-$PYTHON_VER.yml"
+
+gpuci_logger "Activate conda env"
 conda activate dask_image
 
 gpuci_logger "Install cupy"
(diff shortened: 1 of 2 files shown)


continuous_integration/gpuci/build.sh
```diff
@@ -33,16 +33,16 @@ env
...
+
+gpuci_logger "Activate conda env"
 conda activate dask_image
 
 gpuci_logger "Install cupy"
 python -m pip install cupy-cuda112 -f https://pip.cupy.dev/pre
 
-# gpuci_logger "Install dask"
-# python -m pip install git+https://github.com/dask/dask
-
```
jakirkham, line 45: Wonder if we should add the Dask nightly channel ( `dask/label/dev` ) to the environment files used by CI

jakirkham:
Thanks Charles! 🙏

Generally this seems reasonable. Had a somewhat tangentially related question above

continuous_integration/gpuci/build.sh
charlesbluca: Would this change anything in terms of what Dask version is getting picked up? IIUC all the dask-image CI files use a pinned Dask version.

In general I'm okay with unpinning & installing a nightly Dask version as part of GPU CI, though might also be worth exploring if this is something we would like to do for CPU CI as well

jakirkham: Yeah this is a good point. Maybe we can raise an issue and follow up later

Issue #320, opened 2023-06-06
`dask_image.ndmeasure.label` gives different results compared to `scipy.ndimage.label`


vecxoz:
Hi!
I have a binary 3D array and I need to perform connected-component labeling. In my experiments `dask_image` produces different results compared to `scipy`. Another observation is that the number of labels returned by `dask_image` depends on the input `dtype`. Do I actually do everything correctly? Thanks in advance!

```python
import numpy as np # 1.23.5
import scipy # 1.10.1
from dask_image import ndmeasure # 2023.03.0

np.random.seed(0)
X = np.random.choice([0, 1], size=(800, 800, 800), p=[0.7, 0.3]).astype(np.uint8)
structure = scipy.ndimage.generate_binary_structure(3, 3) 

# scipy
label_1, n_labels_1 = scipy.ndimage.label(X, structure)
print(n_labels_1) # 18533

# dask_image
label_2, n_labels_2 = ndmeasure.label(X, structure)
print(n_labels_2.compute()) # 18534

print((label_1 == label_2.compute()).all()) # False
```

**Environment**:

- Dask version: 2023.5.0
- Python version: Python 3.10.10
- Operating System: Ubuntu 20.04.6 LTS
- Install method: pip

jakirkham:
Thanks for the report! 🙏

The labels will be different as designing the algorithm to work in a distributed way required some rethinking

That said, maybe it would be worth computing some kind of similarity matrix between the masks corresponding to each label? Could then threshold this to find the most likely corresponding labels. This might give us a better idea of which labels are related between the two results (as the order may be different) and how far off they are from each other. We could also dig a bit more into the kinds of differences seen between the algorithms

vecxoz:
Hi @jakirkham,
Thank you for your reply. Of course a distributed approach always makes the implementation a bit sophisticated. Btw, I have a question regarding the input size. I found that when the input array is relatively small (e.g. 400x400x400 and smaller) results are identical with `scipy`. But when the input size increases the result starts to deviate. What do you think, is it a coincidence, or there may be some specific relation? Thanks!

jakirkham:
cc @jni (who may also have thoughts here)

jakirkham:
Thanks Marvin! 🙏 Was debating asking as it may make easier to see how things are changing 😄

GenevieveBuckley:
Closed by https://github.com/dask/dask-image/pull/321

PR #321, opened 2023-06-07
ndmeasure.label: fix for arbitrary struct elements

m-albert:
In https://github.com/dask/dask-image/issues/320 @vecxoz showed that the current implementation of `dask_image.ndmeasure.label` is not equivalent to `scipy.ndimage.label` when using a full structuring element. This PR fixes this.

Explanation:

`dask_image.ndmeasure.label` computes label maps for each block independently and then merges these by joining labels that are connected across block boundaries. However, it currently only considers orthogonal boundaries between chunks, which can lead to problems at diagonal chunk borders when using diagonal connectivity structuring elements.

Code changes:

In terms of implementation, `dask_image.ndmeasure.label` considers chunk boundaries by obtaining two-pixel wide slices that connect pairs of neighboring chunks using the function `dask_image.ndmeasure._utils._label._chunk_faces`. This function currently only returns these slices along orthogonal chunk boundaries. This PR extends this function such that the structuring element determines which boundaries are considered. Therefore, this implementation improves the equivalence between the `dask_image` and `scipy.ndimage` implementations while keeping the same dask graph structure for already supported structuring elements.

Tests:

This PR adds a test that documents the identified divergence between the two implementations and addresses the proposed improvement.

@jni and @jakirkham came up with the `dask_image.ndmeasure.label` implementation and could be great reviewers for this :)

### DIFF
--- a/dask_image/ndmeasure/_utils/_label.py
+++ b/dask_image/ndmeasure/_utils/_label.py
@@ -151,6 +151,10 @@ def label_adjacency_graph(labels, structure, nlabels):
         This matrix has value 1 at (i, j) if label i is connected to
         label j in the global volume, 0 everywhere else.
     """
+
+    if structure is None:
+        structure = scipy.ndimage.generate_binary_structure(labels.ndim, 1)
+
     faces = _chunk_faces(labels.chunks, labels.shape, structure)
     all_mappings = [da.empty((2, 0), dtype=LABEL_DTYPE, chunks=1)]
     for face_slice in faces:
--- a/dask_image/ndmeasure/_utils/_label.py
+++ b/dask_image/ndmeasure/_utils/_label.py
@@ -151,7 +151,7 @@ def label_adjacency_graph(labels, structure, nlabels):
         This matrix has value 1 at (i, j) if label i is connected to
         label j in the global volume, 0 everywhere else.
     """
-    faces = _chunk_faces(labels.chunks, labels.shape)
+    faces = _chunk_faces(labels.chunks, labels.shape, structure)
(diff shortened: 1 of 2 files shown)


jni (approved):
Sorry for the slow response here! I think @m-albert's fix is correct and it totally makes sense that we would have forgotten about the higher connectivity case @jakirkham 😅

m-albert:
Thanks for your feedback @jni . Yeah, I figured you might have just not prioritised less common and non-default structuring elements :)

GenevieveBuckley (approved):

GenevieveBuckley:
Thank you @m-albert this is great work 🎉

GenevieveBuckley:
...we should really get a new release out. Anyone up for it?

m-albert:
Thanks for approving and merging @GenevieveBuckley 🎉

> ...we should really get a new release out. Anyone up for it?

I have limited experience with the release process, but thanks to the fantastic [release guide](https://github.com/dask/dask-image/blob/main/docs/release/release_guide.rst) I'd be very happy to give it a shot if you guys agree :)

GenevieveBuckley:
> > ...we should really get a new release out. Anyone up for it?
> 
> I have limited experience with the release process, but thanks to the fantastic [release guide](https://github.com/dask/dask-image/blob/main/docs/release/release_guide.rst) I'd be very happy to give it a shot if you guys agree :)

That would be brilliant, thank you!
And if you find any parts of that guide that could be improved, absolutely feel free to edit that too 😄

jakirkham:
+1

Thanks Marvin! 🙏

Agree with Genevieve. Please share your notes after going through the release process 🙂

m-albert:
Great, will do :)

PR #323, opened 2023-07-03
Fix GitHub actions README badge

dstansby:
I noticed that the CI badge is saying failing, but the CI isn't failing - looks like it just needed updating to the current workflow name. See https://github.com/dstansby/dask-image/blob/patch-1/README.rst for the correctly rendered badge on this branch.


GPUtester:

GenevieveBuckley:
Thanks for noticing, and the quick fix!

Issue #327, opened 2023-07-28
Failing GPU test that isn't run by the gpuCI 


m-albert:
I just ran local tests with cupy installed and got a failing test:

```python
>       out = threshold_local(self.image, block_size, method='gaussian')

tests/test_dask_image/test_ndfilters/test_cupy_threshold.py:37:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
dask_image/ndfilters/_threshold.py:87: in threshold_local
    thresh_image = _gaussian.gaussian_filter(image, sigma, mode=mode,
dask_image/ndfilters/_gaussian.py:65: in gaussian_filter
    sigma = _get_sigmas(image, sigma)
dask_image/ndfilters/_gaussian.py:25: in _get_sigmas
    nsigmas = np.array(sigma)
../../miniconda3/envs/test_dask_image/lib/python3.10/site-packages/dask/array/core.py:1703: in __array__
    x = np.array(x)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.

cupy/_core/core.pyx:1480: TypeError
```
I think the error makes sense, as `sigma` is passed as a cupy backed dask array by `dask_image.ndfilters.threshold_local`:

https://github.com/dask/dask-image/blob/91ea2802a919d3d30198e9473713236ecbc8003e/dask_image/ndfilters/_threshold.py#L83

However, somehow the gpuCI doesn't pick it up.

[...]

PR #329, opened 2023-08-01
Added missing cupy test mark and fixed cupy threshold

m-albert:
This PR fixes https://github.com/dask/dask-image/issues/327 by
- adding the cupy mark to the ignored test
- passing `sigma` in `threshold_local` (gaussian mode) as a numpy array instead of as an array of the same type as the input image.

### DIFF
--- a/dask_image/ndfilters/_threshold.py
+++ b/dask_image/ndfilters/_threshold.py
@@ -80,7 +80,7 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
                                                mode=mode, cval=cval)
     elif method == 'gaussian':
         if param is None:
-            sigma = (da.asarray(block_size, like=image._meta) - 1) / 6.0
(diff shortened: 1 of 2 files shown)


m-albert:
On main, `sigma` which is later passed to `gaussian_filter` is created as a cupy backed dask array when the image is a cupy backed dask array.

```python
sigma = (da.asarray(block_size, like=image._meta) - 1) / 6.0
```

This creates the error

>   ???
E   TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.
cupy/_core/core.pyx:1480: TypeError

I think the fix proposed here makes sense but I might be missing something:

```python
sigma = (np.array(block_size).astype(float) - 1) / 6.0
```

GenevieveBuckley:
Nice catch!

Issue #189, opened 2021-02-08
"wrap" mode for label


rabernat:
Over in https://github.com/scipy/scipy/issues/8218 we are discussing how to implement "wrap" mode (i.e. periodic boundary conditions) for `label` and other morphological functions. (This issue is just about `label`.) We would like to be able to call `label` and have features that cross the array's periodic boundary (in our case, the date line), be counted as a single feature.

I'm opening this issue because
1. If implemented, we would eventually want to propagate this feature downstream to dask-image
2. It strikes me that you have essentially _already implemented it_! Dask-image is able to automatically unify labels across neighbor chunks. This code seems to live in https://github.com/dask/dask-image/blob/main/dask_image/ndmeasure/_utils/_label.py. For "wrap" mode, we could apply basically the same approach, but treating the array as its _own neighbor chunk_.

It sounds like the scipy-devs are open to accepting a PR to implement wrap-mode via a relabeling approach. So I am wondering if you folks here have any advice about how to best extract / adapt your relabeling method to live in scipy itself. Thanks in advance for any advice or hints you might be able to share.

And thanks for the amazing dask-image package! It is super helpful for all our work.

cc @hscannell

jakirkham:
cc @jni

GenevieveBuckley:
@jni authored https://github.com/dask/dask-image/blob/main/dask_image/ndmeasure/_utils/_label.py so I'd say he's probably the best person to guide you.

jakirkham:
cc @grlee77 (also in case you have thoughts on how we might consolidate this upstream)

jni:
Hi everyone,

I'm sorry that I haven't had the bandwidth to respond here properly, much less consider the code carefully. I do love that code, which @jakirkham and I implemented over a feverish couple of days in 2019 here in Melbourne — remember in-person visits?

Anyway, I think your assessment is right @rabernat, the very same code can almost certainly be used to wrap the arrays, modulo some connectivity parameter. It'd be equally fun to implement, should someone have the bandwidth. I'd love to say I would do it but realistically I won't get the chance any time soon.

Holmgren825:
Hey,
This is a bit old now but came across it looking for a way to wrap features over the date line in gridded geodata (Similar use to @rabernat I believe). I've managed to hack together an initial solution which merges features across the boundaries of the array. Would be happy to discuss further and try to get it up to standards if there is still some interest in this.

jakirkham:
Thanks Erik! 🙏

[...]

PR #344, opened 2023-10-11
Wrapping labels over array boundaries

Holmgren825:
I open this to get the discussion started on my hacky implementation of wrapping labels over array boundaries, as mentioned in #189. Currently, it is possible to wrap labels across one, or both, boundaries of a 2d array. This is relatively simple, since one can treat the “wrap” as another face in the `label_adjacency_graph`. However, there are some obvious things that need to be solved before merged:
-  [x] How should wrapping be done in higher dimensions? I guess this could work in the same way for 3d, but nd?
- [ ] Tests are crude. Since there is no reference implementation in SciPy, I've basically just hard coded some arrays now.

There are probably more things to figure out.


GPUtester:

jni:
Hello!

Converting 2D to nD is my bread and butter. 😃 

From an API perspective, rather than a string, we should do (imho), `wrap_axes=None`, and then you can pass in an optional tuple of int for the axes along which you want to wrap.

Then, instead of:

```
    if wrap in ["0", "both"]:
        faces.append(da.hstack([labels[:, [-1]], labels[:, [0]]]))
```

You would do:

```
# lil' helper function
def set_tup_value(tup, idx, value):
    """Return a copy of `tup` with `value` at `idx`."""
    return tuple((elem if i != idx else value) for i, elem in enumerate(tup))

...

for ax in wrap_axes:
    none_slice = (slice(None),) * labels.ndim
    sl_back = set_tup_value(none_slice, ax, [-1])
    sl_front = set_tup_value(none_slice, ax, [0])
    faces.append(da.stack([labels[sl_back], labels[sl_front]], axis=ax)
```

The only downside is that this won't wrap around corners, ie I think the code above suffers from #320 at the corners. That seems like a super tricky fix so I'd personally be happy to punt on that to a later PR, as it seems pretty niche.

As a clue to a fix, I think that *if* you have multiple axes *and* you have a structure with connectivity>1, you would need to examine the wrapped faces from earlier axes and add wrapped faces to those faces. It gets confusing. 😅 But again, I'd do this in a later PR, and perhaps add a known failing test with pytest.mark.xfail for:

```
dask_image.label(
        np.array([[1, 0, 0], [0, 0, 0], [0, 0, 1]]),
        structure=np.ones((3, 3)),
        wrap_axes=(0, 1),
        )
```

Holmgren825:
Awesome, thanks @jni! I threw a `.squeeze`  on `da.stack` since this ended up with an extra dimension, which won't be compatible with the structure.

dask_image/ndmeasure/_utils/_label.py
```diff
@@ -123,7 +123,12 @@ def _to_csr_matrix(i, j, n):
     return mat.tocsr()
 
 
-def label_adjacency_graph(labels, structure, nlabels, wrap=None):
+def set_tup_value(tup, idx, value):
+    """Return a copy of `tup` with `value` at `idx`."""
+    return tuple((elem if i == idx else value) for i, elem in enumerate(tup))
```
Holmgren825, line 128: Note that I changed this to `==`. I guess this comes down to how one thinks about wrapping. As it is now, setting `wrap_axes=(0)` wraps labels across the boundary of the 0th axis, whereas `!=` would wrap the array over the 0th axis and wrap labels across the boundary of the 1st axis.

m-albert:
Thanks for submitting this PR @Holmgren825.

> The only downside is that this won't wrap around corners, ie I think the code above suffers from [#320](https://github.com/dask/dask-image/issues/320) at the corners. That seems like a super tricky fix so I'd personally be happy to punt on that to a later PR, as it seems pretty niche.

@jni I agree that it's a bit of a niche use-case, but I think the logic in `dask_image.ndmeasure._label._chunk_faces` implemented by #321 should already work for the wrapping case (with small modifications).

Here's what I mean: `_chunk_faces` determines the pairs of chunks to consider by applying the given structuring element to an aligned grid of chunks (containing linear chunk indices). 
I.e. it creates a list of chunk faces by iterating over each position in this grid and pairing the current block with the block associated to each entry in the structuring element (in the forward direction).

To support wrap mode, this iteration could simply start at indices -1 instead of 0 for each axis indicated by `wrap_axes`.

Independently of support for connectivity>1, probably it makes sense that all code for determining relevant chunk faces lives in the same place, i.e. currently within `_chunk_faces`.

tests/test_dask_image/test_ndmeasure/test_core.py
```diff
@@ -366,6 +366,97 @@ def test_label(seed, prob, shape, chunks, connectivity):
...
+            np.array([[1, 0, 0], [0, 0, 0], [0, 0, 1]]),
+            (0, 1),
+            marks=pytest.mark.xfail(reason="Can't wrap corner labels"),
+        ),
+    ],
+)
+def test_label_wrap(a, a_res, wrap_axes):
+    d = da.from_array(a, chunks=(5, 5))
+
+    s = np.ones((3, 3))
```
m-albert, line 453: As jni [commented](https://github.com/dask/dask-image/pull/344#issuecomment-1758799104), using structuring elements with connectivity > 1 would lead to problems in the corners. `scipy.ndimage.morphology.generate_binary_structure` is a nice convenience function for creating structuring elements.

m-albert (commented):
@Holmgren825 I left some comments next to the code :)

GenevieveBuckley:
add to allowlist

dask_image/ndmeasure/__init__.py
```diff
@@ -322,6 +322,11 @@ def label(image, structure=None):
             [[0,1,0],
              [1,1,1],
              [0,1,0]]
+    wrap_axes : tuple of int, optional
+        Should labels be wrapped across array boundaries, and if so which axes.
+        - (0) only wrap across the 0th axis.
```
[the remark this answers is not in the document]
Holmgren825: Fixed!

dask_image/ndmeasure/_utils/_label.py
Holmgren825: Yes, I agree. Just wanted to comment it since it differs from the suggestion of @jni.

jni: The suggestion from @jni was pretty abstract and untested and he well could have gotten a sign wrong somewhere 😅 Thanks @Holmgren825 and @m-albert! Sorry the next few weeks are very busy for me so I may not be able to do an in-depth review, but I'll try. Please ping me if there is a rush/you specifically want an extra pair of eyes on something.

m-albert: > To me, indicating the axes over which dask_image.ndmeasure.label should consider the input array to wrap over is most intuitive :) (which is what I think this code does)

I had been wrong earlier in the sense that the code above should be saying `!=` as jni suggested earlier. Because the idea is to only replace the element at index `idx` and leave other elements unchanged (those with `i != idx`).

> Sorry the next few weeks are very busy for me so I may not be able to do an in-depth review, but I'll try. Please ping me if there is a rush/you specifically want an extra pair of eyes on something.

Thanks for your availability @jni :)

Holmgren825: Yes, same here. I guess I got confused thinking about array axes and geographical axes for some reason. Fixed in https://github.com/dask/dask-image/pull/344/commits/3aaeaca8cad1c3523346e82b7f6440a24f5e1388.

Holmgren825:
After some “field” testing on real data, I've further reworked how the `wrap_slices` are generated (which is now takes place in `_chunk_faces`). Since dask arrays don't allow for nd indexing (e.g. https://github.com/dask/dask/issues/4157), the `wrap_slices` are now tuples consisting of slices only. I added a few more tests that should cover some different 3-d use cases:
- Wrapping a single dimension, corners are not connected.
- Wrapping  dim (1, 2), corners are connected.
- Wrapping dim (1, 2), corners are connected, and first and last entry along dim 0 are connected since connectivity > 1.

Maybe these tests are a bit overkill, they add a lot of hard coded arrays, but it is tricky to test the wrapping without it.

m-albert:
Hey @Holmgren825, thank you for addressing the comments above in your follow-up commits. And for adding more tests, which is super useful (also including a 3D example).

> After some “field” testing on real data, I've further reworked how the wrap_slices are generated (which is now takes place in _chunk_faces). Since dask arrays don't allow for nd indexing (e.g. https://github.com/dask/dask/issues/4157), the wrap_slices are now tuples consisting of slices only.

Exactly, I was wondering how to best do this since dask doesn't support fancy indexing yet. I really like your elegant solution of defining slices with a step of `shape[dim]-1`. Super cool :)

> Thanks @m-albert! I agree that it would nicer to move this to _chunk_faces, although I've struggled a bit to understand what's going on in it. One idea I had was that, in the main loop over the blocks, you could stack the bottom block on top when neigh_block[dim] >= numblocks[dim], but these slices do not wrap. So I went with a simpler approach and just moved the loop over the wrap_axes to the end of _chunk_faces, and added a slice that covers the corners of the array. This makes it pass the corner feature test case that previously failed. Lowering the connectivity to one for this case returns two features despite wrapping both axes, which I think is correct.

I completely understand how the code in `_chunk_faces` is a bit tricky to decipher. I tried to comment it a bit better and merged your approach of defining slices with the grid iteration logic in `_chunk_faces` along the lines of what I described [here](https://github.com/dask/dask-image/pull/344#issuecomment-1764340584). Would you agree if I pushed this to this PR branch (`label_wrap`) and we collaborate on it together? Alternatively, we could try to first merge this and I'd open a follow-up PR.

m-albert:
@Holmgren825 Cool, thank you. I pushed the changes. All the tests you had added are passing :)

Next to the discussed logic inside of `_chunk_faces` I joined two for loops and used `yield` instead of `return` in `_chunk_faces` to avoid keeping objects in memory that anyways are consumed right away.

Please feel free to raise things you don't agree with or that should be commented more clearly.

jni:
Amazing! This is a great turnaround speed for this feature — which is quite unique in the Python world??? Like, I don't think you can do this with straight up ndimage??? How cool is that??? 👏

m-albert:
Very exciting indeed 😁

I think reviews are welcome at this point.

Generally speaking, I guess for most of its functionality `dask-image` opens up `scipy.ndimage` for dask arrays by sticking to the reference API as closely as possible (often implementing a subset of the functionality). However here there's an additional feature introduced. Which in this case might actually not only be relevant in the context of dask arrays but also beyond as discussed [here](https://github.com/dask/dask-image/issues/189#issue-803625484) and [here](https://github.com/scipy/scipy/issues/8218). Is there a risk of scope creep? I guess a good reason for adding this feature here is that it's enabled by the chunk logic used and implemented by the same `dask-image.ndmeasure.label` functionality.

m-albert:
Gentle ping @jakirkham @GenevieveBuckley. Would you have opinions or review comments on this PR?

jni:
Just writing to say I have a strong opinion that this should go in. 😂

m-albert:
Time to take some action here :)

I think this PR is ready. In case there are no additional comments or objections in the next days, I'd merge this later this week.

m-albert:
Merged 🎉

Thank you @Holmgren825 for your contribution to dask-image and implementing this new feature that extends functionality in `scipy.ndimage`!

Thanks to @jni for the groundwork, reviewing and cheerful comments!

Thanks also to @rabernat for starting the [conversation](https://github.com/dask/dask-image/issues/189) on this.

dask_image/ndmeasure/_utils/_label.py
```diff
@@ -202,44 +217,59 @@ def _chunk_faces(chunks, shape, structure):
...
             neigh_block = [curr_block[dim] + pos_structure_coord[dim] - 1
                            for dim in range(ndim)]
 
-            if max([neigh_block[dim] >= numblocks[dim] for dim in range(ndim)]): continue
+            if max([neigh_block[dim] >= block_summary.shape[dim]
+                    for dim in range(ndim)]): continue
 
             # get neighbor slice index
-            ind_neigh_block = block_summary[tuple(neigh_block)]
+            ind_curr_block = block_summary[tuple(curr_block)]
```
jakirkham, line 256: Added a small change to this comment in PR: https://github.com/dask/dask-image/pull/353

jakirkham:
Thanks Marvin for shepherding this in!

Also thanks Juan for reviewing!

And of course thanks Erik for reaching out and sharing this contribution with us

Hopefully it helps others with similar needs 🙂

Issue #352, opened 2024-02-23
New release


m-albert:
Probably we should create a release given the new functionality from https://github.com/dask/dask-image/pull/213 and https://github.com/dask/dask-image/pull/344 !

Happy to do this in the next days if you guys agree @jakirkham @GenevieveBuckley.

jakirkham:
Would suggest we include PR: https://github.com/dask/dask-image/pull/354

That way users can see the new features added in the docs associated with the new version

Otherwise +1 on releasing. Thanks for handling this Marvin! 🙏

jakirkham:
Think we should be good to go

GenevieveBuckley:
Absolutely!

m-albert:
Great :) Unless one of you guys would do it earlier, I'd plan to do it on Monday.

ericpre:
xref https://github.com/conda-forge/dask-image-feedstock/pull/20, the current release doesn't build on conda-forge:

```

2024-03-10T13:47:56.5768709Z Processing $SRC_DIR
2024-03-10T13:47:56.5828345Z   Added file://$SRC_DIR to build tracker '/tmp/pip-build-tracker-1mw32res'
2024-03-10T13:47:56.5829325Z   Running setup.py (path:$SRC_DIR/setup.py) egg_info for package from file://$SRC_DIR
2024-03-10T13:47:56.5829744Z   Created temporary directory: /tmp/pip-pip-egg-info-4o8laagz
2024-03-10T13:47:56.5829985Z   Preparing metadata (setup.py): started
2024-03-10T13:47:56.5830218Z   Running command python setup.py egg_info
2024-03-10T13:47:56.7345781Z   /home/conda/feedstock_root/build_artifacts/dask-image_1710078359865/work/versioneer.py:420: SyntaxWarning: invalid escape sequence '\s'
2024-03-10T13:47:56.7356532Z     LONG_VERSION_PY['git'] = '''
2024-03-10T13:47:56.7449342Z   Traceback (most recent call last):
2024-03-10T13:47:56.7461702Z     File "<string>", line 2, in <module>
2024-03-10T13:47:56.7471938Z     File "<pip-setuptools-caller>", line 34, in <module>
2024-03-10T13:47:56.7483120Z     File "/home/conda/feedstock_root/build_artifacts/dask-image_1710078359865/work/setup.py", line 56, in <module>
2024-03-10T13:47:56.7493465Z       version=versioneer.get_version(),
[...]

PR #306, opened 2023-03-29
Switch to pyproject.toml package setup

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/305

PR #357, opened 2024-03-18
Update versioneer to version 0.29 for compatibility with python 3.12

m-albert:
@ericpre [reported](https://github.com/dask/dask-image/issues/352) that the current release doesn't build on conda-forge.

The problem is that the version of `versioneer` used in `dask-image` is incompatible with python=3.12 (see e.g. https://github.com/warner/python-spake2/issues/13.)

This PR updates versioneer to version 0.29.


jakirkham:
Thanks Marvin! 🙏

Restarted CI as one of the gpuCI builds had an error

We may also want to consider getting PR ( https://github.com/dask/dask-image/pull/306 ) in to use `pyproject.toml`. May still need a bit more work

jakirkham (approved):

Issue #362, opened 2024-04-22
Tests failing with python 3.11.9: TypeError during import of dask.dataframe


m-albert:
https://github.com/dask/dask-image/pull/361 reveals that tests are currently failing with python 3.11.9.

This seems to be a known problem with dask (https://github.com/pycaret/pycaret/issues/3960#issuecomment-2045881422, https://github.com/dask/dask/pull/11035) that is fixed by using `dask>=2024.4.1`.

PR #363, opened 2024-04-22
Pin dask to 2024.4.1 to avoid error during dask.dataframe import with python 3.11.9

m-albert:
Fixes https://github.com/dask/dask-image/issues/362, see https://github.com/dask/dask/pull/11035.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,8 +29,8 @@ with open("HISTORY.rst") as history_file:
     history = history_file.read()
 
 requirements = [
-    "dask[array] >=2023.2.0",
-    "dask[dataframe] >=2023.2.0",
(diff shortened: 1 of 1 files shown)


m-albert:
Not entirely sure whether pinning the very recent dask>=2024.4.1 is the right approach here. Would be grateful for a review @jakirkham :) 🙏

GenevieveBuckley:
I thought for a bit that only dask=2024.4.0 was affected, but I see the same error pop up when I try earlier versions as well. I think you're right, and it's safest to require dask>=2024.4.1

m-albert:
Thank you for looking into this!

PR #366, opened 2024-05-13
Maintenance: fix CI test errors

GenevieveBuckley:
Code maintenance, fixing some errors that have popped up in the CI test suite. Mostly adjustments for code deprecation, and minor API changes in other libraries.

PR #367, opened 2024-05-14
Update CI test environments

GenevieveBuckley:
This PR:
* Updates the python 3.9 CI test environment, with a configuration I've checked works on arm64 Mac.
* Updates all of the environment yaml files to use dask==2024.4.1 (see https://github.com/dask/dask-image/pull/363)

I didn't think we needed to do this, since all of the CI checks for https://github.com/dask/dask-image/pull/363 passed before we merged it, but manually triggering CI on the main branch is failing ([see here](https://github.com/dask/dask-image/actions/runs/9074896375)). So this is an attempt to fix that problem.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      # fail-fast: true
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
         python-version: ["3.9", "3.10", "3.11"]
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
-      # fail-fast: true
+      fail-fast: false
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
         python-version: ["3.9", "3.10", "3.11"]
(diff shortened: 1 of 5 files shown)


jakirkham (approved):

.github/workflows/test_and_deploy.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      # fail-fast: true
```
GenevieveBuckley, line 16:
```suggestion
      fail-fast: true
```

GenevieveBuckley: We need to wait for the CI checks to finish running, and see if they all successfully  create conda environments. Then we can commit this suggestion, and merge the PR.

Note: we are still expecting to see pytest failures here, which will be fixed by https://github.com/dask/dask-image/pull/366. But if pytest runs at all, then the conda environment definitely was created properly.

GenevieveBuckley:
We need to wait for the CI checks to finish running, and see if they all successfully  create conda environments. Then we can commit this suggestion, and merge the PR.

Note: we are still expecting to see pytest failures here, which will be fixed by https://github.com/dask/dask-image/pull/366. But if pytest runs at all, then the conda environment definitely was created properly.

.github/workflows/test_and_deploy.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   test:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
GenevieveBuckley, line 16:
```suggestion
      fail-fast: true
```

Issue #358, opened 2024-03-18
Add Python 3.12


jakirkham:
Opening this issue to track adding Python 3.12 support.

Some work was already done in PR: https://github.com/dask/dask-image/pull/357

Though more work may still be needed

Also would be good to include Python 3.12 on CI

Issue #369, opened 2024-05-14
GPU CI test errors 


GenevieveBuckley:
**Describe the issue**:
I've noticed the GPU CI job has been failing.

Link to GPU CI:
https://gpuci.gpuopenanalytics.com/job/dask/job/dask-image/job/prb/job/dask-image-prb/

GenevieveBuckley:
I'm seeing errors more frequently for the python 3.9 GPU CI build. I hope I didn't cause this by changing the python 3.9 environment file in PR https://github.com/dask/dask-image/pull/367. That's one thing to check.

GenevieveBuckley:
Also, possibly we should update which versions of python are used for the GPU tests (especially given https://github.com/dask/dask-image/issues/358, we don't want the GPU tests falling too far out of date)

GenevieveBuckley:
Closed by https://github.com/dask/dask-image/pull/368

PR #370, opened 2024-05-14
Add python 3.12 support

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/358


GenevieveBuckley:
This PR does not update the GPU CI test matrix (that still has python 3.9 and 3.10 only, nothing more recent). Probably we should handle that in a separate PR (also see https://github.com/dask/dask-image/issues/369)

PR #372, opened 2024-05-15
Bump GPU CI highest python version to 3.12

GenevieveBuckley:
Now that we have support for python 3.12 https://github.com/dask/dask-image/pull/370, it's a good idea to bump the highest GPU CI job to also use python 3.12 (don't want to let it get too far out of date)


jakirkham:
Thanks Genevieve! 🙏

Agree this is a good idea

We may need a change in the CI images. Asked about that in issue: https://github.com/rapidsai/dask-build-environment/issues/93

Also thank you for all of your hard work getting CI into good shape here 🙂

continuous_integration/gpuci/axis.yaml
```diff
@@ -2,6 +2,7 @@ PYTHON_VER:
 - "3.9"
 - "3.10"
 - "3.11"
+- "3.12"
```
GenevieveBuckley, line 5: Previously we've had just two python versions running the GPU tests (lowest & highest available versions, currently that's python 3.9 and 3.12). Do we want to remove these intermediate python versions, and keep testing only against two python versions? If so, we'll edit these lines.

Issue #17, opened 2018-08-30
Drop slicerator CI install hack (once fixed)


jakirkham:
_From @jakirkham on June 7, 2017 20:3_

We had to add a hack for the CIs in commit ( https://github.com/dask-image/dask-imread/commit/ffd6cc5bb40fa24d7652520d0598fa71b87331e4 ) to `pip` install `slicerator` in addition to `conda` installing it as `pip` could not recognize that it was already installed. This seems to be some consequence of the choice to use Python modules and not a Python package to hold the library content.

Have proposed a fix upstream in PR ( https://github.com/soft-matter/slicerator/pull/26 ), which seems to resolve the issue. Once upstream merges that fix and releases, we should drop this hack. Though admittedly that will hinge on dropping Python 3.4 as we won't have a new Python 3.4 package to work with from `conda-forge`.

_Copied from original issue: dask-image/dask-imread#7_

jakirkham:
Some discussion on the new `slicerator` release in issue ( https://github.com/soft-matter/slicerator/issues/35 ).

jakirkham:
This should be possible now that there is a 1.0.0 release.

GenevieveBuckley:
This issue can be closed now. 
The minimum version of slicerator used anywhere in this repo is v1.0.0, and I can't see this install hack anywhere either (probably was removed a long time ago)

PR #374, opened 2024-05-17
Update minimum versions in `dependencies`

jakirkham:
It looks like a few dependencies had higher minimum versions when using `setup.py`

https://github.com/dask/dask-image/blob/243fb0a5de55f5bc5b8650985d26381a7c22889e/setup.py#L31-L39

Though we wound up with older versions after switching to `pyproject.toml`

https://github.com/dask/dask-image/blob/49d9c336bc4551bbfc89c7d3527b0c176ae6f4a5/pyproject.toml#L26-L35

Guessing this is just because the `pyproject.toml` PR has fallen out of date in this area and we missed it

So went ahead and bumped these. Though please let me know if I'm missing something

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,11 +26,9 @@ classifiers = [
 dependencies = [
     "dask[array,dataframe] >=2024.4.1",
     "numpy >=1.18",
-    "scikit-image >=0.19.3",
     "scipy >=1.7.0",
     "pandas >=2.0.0",
     "pims >=0.4.1",
-    "slicerator >=1.1.0",
     "tifffile >=2018.10.18",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "scipy >=1.7.0",
     "pandas >=2.0.0",
     "pims >=0.4.1",
-    "slicerator >= 1.1.0",
+    "slicerator >=1.1.0",
     "tifffile >=2018.10.18",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,9 +25,9 @@ classifiers = [
 ]
 dependencies = [
     "dask[array,dataframe] >=2024.4.1",
-    "numpy >=1.11.3",
+    "numpy >=1.18",
     "scikit-image >=0.19.3",
-    "scipy >=0.19.1",
+    "scipy >=1.7.0",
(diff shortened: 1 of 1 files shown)


jakirkham:
Also assuming `scikit-image` and `slicerator` were added intentionally. Please let me know if not and can update accordingly

jakirkham:
@GenevieveBuckley could you please review?

GenevieveBuckley (approved):

GenevieveBuckley:
> Guessing this is just because the pyproject.toml PR has fallen out of date in this area and we missed it
> So went ahead and bumped these. Though please let me know if I'm missing something

Yes, I think the PR just fell out of date between it being opened and merged.

> Also assuming scikit-image and slicerator were added intentionally. Please let me know if not and can update accordingly

I added them because they are included in the environment yaml files. I was trying to make sure it wouldn't matter which method someone used to install it, you'd still get something that matches. That might not have been the best thing to do though.

I also thought for some reason that scikit-image was required for a new-ish function, but I can't actually see it being imported anywhere (including the tests). So maybe it shouldn't be there at all.

I think slicerator is a dependency of one of our other dependencies, so it might not be a thing we need to worry about in the pyproject.toml at all. Like I said, I was just trying to match the environment files, but that's maybe not the best strategy.

GenevieveBuckley:
I also saw this issue from Mark about switching from dask to dask-core: https://github.com/conda-forge/dask-image-feedstock/issues/16

jakirkham:
No worries. It's better to move forward with changes (like `pyproject.toml`) and fix a few things afterwards than let it linger indefinitely. So think you did the right thing 🙂

Think `scikit-image` was added for testing originally. If that is all we are using it for still, it should be safe to drop. So went ahead and did that. If we want to use things from `scikit-image` in the future, we can always add it back then

Regarding `slicerator` that was added as a workaround ( https://github.com/dask/dask-image/issues/17 ). Think we are safe to drop it at this point. So went ahead and did that

The `dask-core` bit is only relevant for the conda-forge feedstock. So let's save that discussion for there

Maybe there is a better way to manage the dependencies here and call out what is needed for testing to make it clearer. Open to thoughts on that, but maybe we can discuss in a new issue?

GenevieveBuckley:
Thank you for doing this @jakirkham

jakirkham:
Happy to help 🙂

PR #377, opened 2024-05-21
Install dask-image with `pip` on RTD

jakirkham:
* Install on RTD with `pip` (keeps the git repo clean, which likely causes the ".dev0" version issue)
* Disable `setuptools-scm`'s `version_scheme` guess (states versions in terms of distance passed the old version)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,3 +10,8 @@ sphinx:
 
 conda:
   environment: continuous_integration/environment-doc.yml
+
+python:
+  install:
+    - method: pip
+      path: .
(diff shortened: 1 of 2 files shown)


GenevieveBuckley:
Looks good, happy for us to give it a go and see if that fixes things.

The only other relevant part of the code base is the `docs/conf.py` lines:
https://github.com/dask/dask-image/blob/main/docs/conf.py#L70-L73

I don't know whether RTD is pulling in `version` or `release` and don't know how to find out, but that's one other thing to check if merging this PR doesn't help.

GenevieveBuckley (approved):

jakirkham:
Sounds good. Let's see what happens. Will do another tag to test. Hopefully we can iron out these remaining release hiccups together 🙂

Issue #375, opened 2024-05-17
Stable docs page shows development version


dstansby:
At https://image.dask.org/en/stable/, there's a non-stable version shown:

<img width="278" alt="Screenshot 2024-05-17 at 11 22 43" src="https://github.com/dask/dask-image/assets/6197628/4993916c-bcc9-4f5d-9013-9e1ff40aa90e">

I'm guessing this should be a stable version?

GenevieveBuckley:
This might be because [ReadTheDocs expects ](https://docs.readthedocs.io/en/stable/versions.html#how-we-envision-versions-working) ([like these examples](https://packaging.python.org/en/latest/specifications/version-specifiers/#examples-of-compliant-version-schemes)), and the [dask-image version tags](https://github.com/dask/dask-image/tags) all start with the letter `v`.

GenevieveBuckley:
> This might be because [ReadTheDocs expects ](https://docs.readthedocs.io/en/stable/versions.html#how-we-envision-versions-working) ([like these examples](https://packaging.python.org/en/latest/specifications/version-specifiers/#examples-of-compliant-version-schemes)), and the [dask-image version tags](https://github.com/dask/dask-image/tags) all start with the letter `v`.

This can probably be solved by using [a custom match](https://docs.readthedocs.io/en/stable/automation-rules.html#custom-matches) in the [automation rules](https://blog.readthedocs.com/automation-rules/) for ReadTheDocs.

@jakirkham will probably have to do this: it looks like you have to log in to ReadTheDocs to both (1) set automation rules, and (2) activate any tags ReadTheDocs has not previously identified as a version tag. 

I tried logging in to ReadtheDocs with github, but I don't see anything related to dask-image linked to my account.

GenevieveBuckley:
Separate to this discussion, the main docs page at https://image.dask.org does redirect to https://image.dask.org/en/latest/ (not stable). That is deliberate, from my understanding (mostly because we often have a delay between a new feature going in and an official release being made. It seems better if people know a new thing exists rather than not know)

jakirkham:
> I tried logging in to ReadtheDocs with github, but I don't see anything related to dask-image linked to my account.

@GenevieveBuckley What is your handle on RTD?

@m-albert would you like access to RTD as well? If so, what is your handle?

GenevieveBuckley:
> @GenevieveBuckley What is your handle on RTD?

It's GenevieveBuckley
[...]

PR #379, opened 2024-05-21
ReadTheDocs: fix displayed version number in top left corner

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/375

Fix problem where RTD displays full release version string (eg: X.Y.Z.dev0+gXXXXXX)., not shorter X.Y.Z version


GenevieveBuckley:
It works!
Screenshot 2024-05-22 at 10 07 44 am

jakirkham:
Thanks Genevieve! 🙏

Am wondering if we should somehow indicate to users that the release they are looking at is under development (in comparison to a tagged release). What do you think?

PR #380, opened 2024-05-22
Display dev version numbers on ReadTheDocs latest

GenevieveBuckley:
https://github.com/dask/dask-image/pull/379 may have fixed https://github.com/dask/dask-image/issues/375, but possibly too aggressively - the truncated short version number might be shown even for the development branch at ReadTheDocs latest/


docs/conf.py
```diff
@@ -71,6 +71,10 @@
 release = dask_image._version.__version__
 # The short X.Y.Z version.
 version = '.'.join(release.split('.')[:3])
+if "dev" in release:
+    display_version = release
```
jakirkham, line 75:
```suggestion
    display_version = "(under development)"
```

jakirkham (commented):
Thanks Genevieve! 🙏

Think you are on to something

Wonder to what extent the `setuptools-scm` development versions are meaningful to our users

Maybe we can do something like this?

Also no strong feelings about this approach if you have other ideas 🙂

docs/conf.py
GenevieveBuckley: I've changed your suggestion from `"(under development)"` -> `"(development version)"`.

It might just be me, but "under development" makes me think of those old "under construction" website pages that suggest the website is unfinished.

GenevieveBuckley:
Sure, we can do a heading like "dask-image (development version) documentation"

I don't have strong feeling about it either, except that whatever it shows should be correct information. That seems like a fine way to handle it.

docs/conf.py
jakirkham: Sounds reasonable 🙂

jakirkham (approved):
Thanks Genevieve! 🙏

PR #383, opened 2024-07-17
Update conf.py, sphinx context injection deprecated in ReadTheDocs

GenevieveBuckley:
Closes https://github.com/dask/dask-image/issues/382

ReadTheDocs is deprecating sphinx context injection. We need to set the canonical url a different way.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -123,9 +123,10 @@ exclude_patterns = ['_build']
 # Set canonical URL from the Read the Docs Domain
 html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
 
-html_context = {}
 # Tell Jinja2 templates the build is running on Read the Docs
 if os.environ.get("READTHEDOCS", "") == "True":
+    if "html_context" not in globals():
+        html_context = {}
     html_context["READTHEDOCS"] = True
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -120,6 +120,13 @@ exclude_patterns = ['_build']
 
 # -- Options for HTML output -------------------------------------------
 
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
+
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
+    html_context["READTHEDOCS"] = True
+
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 html_theme = 'dask_sphinx_theme'
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -123,6 +123,7 @@ exclude_patterns = ['_build']
 # Set canonical URL from the Read the Docs Domain
 html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
 
+html_context = {}
 # Tell Jinja2 templates the build is running on Read the Docs
 if os.environ.get("READTHEDOCS", "") == "True":
     html_context["READTHEDOCS"] = True


GenevieveBuckley:
Right now the code change to `conf.py` is a straight copy-paste from the [suggested code snippet](https://github.com/dask/dask-image/issues/382). 

Questions:
* Are there any sphinx extensions in a dependency list somewhere that we need to remove?
* Should I change this line to not use an empty string as the default, but instead show `image.dask.org`, like this?
```python
html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "image.dask.org")
```

m-albert:
Don't have any input here for now but I'm linking https://github.com/dask/dask/issues/11229.

docs/conf.py
```diff
@@ -120,6 +120,13 @@
 
 # -- Options for HTML output -------------------------------------------
 
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
+
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
+    html_context["READTHEDOCS"] = True
```
jakirkham, line 128: Do we need to define `html_context`? Looks like the RTD build doesn't find it currently

```suggestion
html_context = {}
# Tell Jinja2 templates the build is running on Read the Docs
if os.environ.get("READTHEDOCS", "") == "True":
    html_context["READTHEDOCS"] = True
```

GenevieveBuckley:
Better now? https://dask-image--383.org.readthedocs.build/en/383/api.html

jakirkham:
Oops sorry missed your comment Genevieve 😞

Given it passes RTD, yes! 🎉

jakirkham:
rerun tests

docs/conf.py
```diff
@@ -120,6 +120,14 @@
 
 # -- Options for HTML output -------------------------------------------
 
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
+
+html_context = {}
+# Tell Jinja2 templates the build is running on Read the Docs
+if os.environ.get("READTHEDOCS", "") == "True":
+    html_context["READTHEDOCS"] = True
```
jakirkham, line 129: Looking at [the RTD blogpost]( https://about.readthedocs.com/blog/2024/07/addons-by-default/ ), guess we need this

```suggestion
# Tell Jinja2 templates the build is running on Read the Docs
if os.environ.get("READTHEDOCS", "") == "True":
    if "html_context" not in globals():
        html_context = {}
    html_context["READTHEDOCS"] = True
```

Sorry didn't have a good grasp on what was happening there (and honestly still don't). Guessing there is some RTD magic happening under-the-hood

docs/conf.py
```diff
@@ -120,6 +120,14 @@
 
 # -- Options for HTML output -------------------------------------------
 
+# Set canonical URL from the Read the Docs Domain
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "")
```
jakirkham, line 124: > * Should I change this line to not use an empty string as the default, but instead show `image.dask.org`, like this?
>
> ```python
> html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "image.dask.org")
> ```


After reading [this doc page]( https://docs.readthedocs.io/en/stable/canonical-urls.html ), it sounds like RTD checks the domains listed in the RTD settings, which is just `image.dask.org` (as seen below)

Screenshot 2024-09-24 at 9 35 39 PM

Then passes it in as the environment variable `READTHEDOCS_CANONICAL_URL`, which we pick up on the code line above

While we could change the fallback default, it is sounds like

1. Sounds like RTD does the right for us
2. Having a default that smokes out any issues (like empty string does) is useful
3. DRY

Think for these reasons having the setting in one place should work for us and keep maintenance straightforward

Though do appreciate this question and the exercise of thinking through it was helpful

jakirkham:
> * Are there any sphinx extensions in a dependency list somewhere that we need to remove?

There is this list of dependencies

https://github.com/dask/dask-image/blob/77a29ba78db608c5fe21f9f51745ec4bd01d5681/continuous_integration/environment-doc.yml#L20-L30

Also this list of Sphinx extensions

https://github.com/dask/dask-image/blob/77a29ba78db608c5fe21f9f51745ec4bd01d5681/docs/conf.py#L41-L48

Perhaps we want to revisit the former (later). That looks unrelated

The latter all come from Sphinx. So think we can leave them as-is

The blogpost also mentions some new features like [`flyout_display`]( https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html#confval-flyout_display ) and [addons]( https://docs.readthedocs.io/en/stable/addons.html#enabling-read-the-docs-addons )

Personally am more interested in keeping the existing docs functional than adding new bells and whistles, but could be convinced otherwise if there are compelling needs/interests in enabling these

jakirkham (commented):
Thanks Genevieve! 🙏

After reading the blogpost more closely, realize I goofed in my last suggestion. Included a new one based on my re-reading. Though please double check

Also tried to answer your other questions. Please let me know whether that is helpful. Happy to discuss anything further as needed

Thanks again for picking this up 🙂

jakirkham:
Pushed a small fix for `html_context` based on [this RTD blogpost]( https://about.readthedocs.com/blog/2024/07/addons-by-default/ )

jakirkham (approved):
Thanks Genevieve! 🙏

Approving as I think this is all we need. Though please double check

If we do find something needs a change post-merge, we can always do another PR

GenevieveBuckley:
Thanks John!

Issue #335, opened 2023-08-03
New Dask Arrow-based strings cause test failures


jakirkham:
(Edited by @m-albert)

In the presence of `pyarrow`, dask by default assumes dataframes of type object to be pyarrow strings (see https://github.com/dask/dask/issues/10139#issuecomment-1655928619).

This creates problems revealed by failing tests (e.g. `test_dask_image/test_ndmeasure/test_find_objects.py::test_3d_find_objects`)

https://github.com/dask/dask-image/blob/67540af25597f84e4a642d644ba30dce7aebe753/dask_image/ndmeasure/_utils/_find_objects.py#L68-L70

`dd.from_delayed(df1, meta=meta).compute().dtypes`

Working install:
> 0    object
1    object
2    object
dtype: object

Failing install:
>0    string[pyarrow]
1    string[pyarrow]
2    string[pyarrow]
dtype: object

The failing test had come up when releasing v2023.08.0 in https://github.com/conda-forge/dask-image-feedstock/pull/14.

@jakirkham found that `pyarrow` is installed with the [conda distribution of dask](https://github.com/conda-forge/dask-feedstock/blob/89ff1ea9116e0798d9073d819a6b9b22fa1d0917/recipe/meta.yaml#L31), but not when [installing over pip](https://github.com/dask/dask/blob/main/pyproject.toml), where it just part of the `[complete]` target.

Also @jakirkham found that the above described conflicting behaviour can be [turned off using the dask configuration](https://github.com/dask-contrib/dask-sql/commit/af180620216793b7c6125b7db8f98dd0bba038de#diff-a31c7ed5d35f5ed8233994868c54d625b18e6bacb6794344c4531e62bd9dde59R15-R16).

He did this for the tests performed by the dask-image conda feedstock on [v2023.08.0](https://github.com/conda-forge/dask-image-feedstock/pull/14).

jakirkham:
Please fill free to edit and fill this issue out Marvin 🙂

Just needed a placeholder for tracking

jakirkham:
Would also be good to make a note in issue (where feedback is being collected): https://github.com/dask/dask/issues/10139

Ideally with a simple reproducer

m-albert:
Despite the passing tests, potentially users who installed dask-image over conda would still experience the above described problem when using `ndmeasure.find_objects` (need to check whether there's more affected).

[...]

PR #384, opened 2024-07-18
fix KeyError: "None of [Index(['0_x', '1_x', '0_y', '1_y'], dtype='object')] are in the [columns]" in find_objects

joshua-gould:


m-albert:
@joshua-gould Thanks for your PR!

I was trying to reproduce the error you're fixing and found that it relates to https://github.com/dask/dask-image/issues/335.

The following code
```python
    test_labels = da.zeros((10, 10), dtype='int', chunks=(3, 3))
    test_labels[0, 0] = 1
    computed_result = dask_image.ndmeasure.find_objects(test_labels).compute()
```


fails in the presence of `pyarrow` in the environment and runs through in its absence.

jmuhlich:
I ran into this issue with find_objects too, and setting `dataframe.convert-string` to False does _not_ fully fix it for me. (`dask=2025.2.0`, `dask-image=2024.5.3`, also tested with dask-image latest from github today)  I found that triggering the issue depends on the specific layout of the image and which chunks are all zero. Generally it seems that empty chunks "earlier" (left or above?) are problematic. A fully empty label image that has more than one chunk in each dimension always errors. The only thing that fixes it for me is reverting to dask=2024.12.1 without dask-expr.

m-albert:
Thanks @jmuhlich for reporting this here!

It seems that the following example:

```python
import dask.array as da
import dask_image.ndmeasure

test_labels = da.zeros((10, 10), dtype='int', chunks=(3, 3))
test_labels[0, 0] = 1
computed_result = dask_image.ndmeasure.find_objects(test_labels).compute(scheduler='single-threaded')
```

1) fails on main
2) fails on https://github.com/dask/dask-image/pull/393 (which sets dataframe.convert-string to False)
3) runs through on this PR

In this PR, @joshua-gould works around problems that occur when merging dask dataframes.

Also [here](https://github.com/dask/dask/blob/df372bdcdac8e7a7ba7416501bec4dff85e2b650/dask/dataframe/multi.py#L133-L143) theres a mention of a pandas bug when merging dataframes. The error here might be related to that.

I didn't have the time yet to find out what's going wrong in the merge. I think it'd be good to report the results of this upstream.

Independent of upstream we should incorporate this workaround here I think.

> I found that triggering the issue depends on the specific layout of the image and which chunks are all zero

@jmuhlich Does the code in this PR fix the problems you mention?

jakirkham:
Fixed up some conflicts introduced by a recent PR fixing CI: https://github.com/dask/dask-image/pull/393

Hope that is ok

Please feel free to tweak further as needed

m-albert:
## Quick summary

This PR fixes `find_objects` for a straight forward use case, and tests for it.

### Problem on main

The following fails

```python
import dask.array as da
import dask_image.ndmeasure

test_labels = da.zeros((10, 10), dtype='int', chunks=(3, 3))
test_labels[0, 0] = 1
computed_result = dask_image.ndmeasure.find_objects(test_labels).compute(scheduler='single-threaded')
```

### The fix in this PR

The following fails with empty `df1` and `df2` in some cases:

```python
ddf = dd.merge(df1, df2, how="outer", left_index=True, right_index=True)
```

The following workaround in this PR fixes it:

```python
    if len(df1) > 0 and len(df2) > 0:
        ddf = dd.merge(
            df1, df2,
            how="outer", left_index=True, right_index=True)
    elif len(df1) > 0:
        ddf = df1
    elif len(df2) > 0:
        ddf = df2
    else:
        ddf = pd.DataFrame()
```

### Conclusion

I suspect there's a problem with `dd.merge` when working on empty frames, however I couldn't find a minimal reproducer yet.

Since the workaround here works and fixes a problem that came up for different people, I'd propose to merge this PR as is and report a potential problem upstream once we identify it.

What do you think @jakirkham? 🙏

m-albert:
This came up in a new issue again https://github.com/dask/dask-image/issues/403 so I'll go ahead and merge this.

If further discussion / changes are required I suggest we open a new issue! Thanks @joshua-gould for this fix and everyone for contributing 🙏

Issue #199, opened 2021-03-31
Idea: fully parallel ndmeasure.label


jni:
We implemented distributed labeling in #94.

@chrisroat mentions [here](https://github.com/dask/dask-blog/issues/47#issuecomment-809924075) that instead of adding the total number of labels from previous chunks, we can uniquify each block's labels by using the top 32 bits of a 64-bit integer, assuming neither the labels of a single chunk nor the total number of chunks exceed 32 bits, which is a pretty safe assumption.

The issue with this approach is that you end up with very large label ids, which don't play well with some of the downstream processing, which requires creating arrays of size `max_label`.

@jakirkham [came up](https://github.com/dask/dask-blog/issues/47#issuecomment-811444438) with an interesting solution: instead of using the top 32 bits, use the bottom n bits, where n is large enough to accommodate the number of blocks. Since the number of blocks is known at the start, we can bit-shift the labels by that amount and do this fully in parallel. This should result in at most 2x "wastage" in label space, which is probably acceptable for the benefit of a cleaner, more parallel task graph.

PR #409, opened 2025-09-09
Add label merging function (and improve ndmeasure.label)

m-albert:
## PR

This PR contains work I've been wanting to discuss and push here for a while.

Highlights:
1. Improvements to `ndmeasure.label`:
    1. Implementation of the idea by @jakirkham and @jni to make labels unique by using the lower and upper bits of the label array (see [here](https://github.com/dask/dask-image/issues/199)). This improves parallelism at the expensive of returning non-contiguous labels.
    1. The current implementation of `ndmeasure.label` doesn't output contiguous (sequential) labels (also before implementing the point above). This PR adds a parameter `produce_sequential_labels` to `dask_image.ndmeasure.label` to output contiguous labels.
1. Implementation of a chunked version of `skimage.segmentation.relabel_sequential`: `ndmeasure.relabel_sequential` (this provides the functionality e.g. of the point above, happy to discuss whether the function should be public)
1. Implementation of `ndmeasure.merge_labels_across_chunk_boundaries` (details below)
    1. Use case 1: merge touching labels
    2. Use case 2: merge labels above a given IoU within the overlap.

## merge_labels_across_chunk_boundaries
What's the idea behind `ndmeasure.merge_labels_across_chunk_boundaries`?

### Visual abstract

Cases of overlap = 0 (top) and overlap > 0 (bottom)

image

### Description

From the docstring:

```python
def merge_labels_across_chunk_boundaries(
        labels,
        overlap_depth=0,
        iou_threshold=0.8,
        structure=None,
        wrap_axes=None,
        trim_overlap=True,
        produce_sequential_labels=True,
        ):
    """
    Merge labels across chunk boundaries.

    Each chunk in ``labels`` has been labeled independently, and the labels
    in different chunks may overlap. This function tries to merge labels across
    chunk boundaries using a strategy dependent on ``overlap``:
    - If ``overlap > 0``, the overlap region between chunks is used to
      determine which between each pair of chunks should be merged.
    - If ``overlap == 0``, labels that touch across chunk boundaries are merged.

    Parameters
    ----------
    labels : dask array of int
        The input labeled array, where each chunk is independently labeled.
    overlap_depth : int, optional
        The size of the overlap region between chunks, e.g. as produced by
        `dask.array.overlap` or `map_overlap`. Default is 0.
    iou_threshold : float, optional
        If ``overlap > 0``, the intersection-over-union (IoU) between labels
        in the overlap region is used to determine which labels should be
        merged. If the IoU between two labels is greater than ``iou_threshold``,
        they are merged. Default is 0.8.
    structure : array of bool, optional
        Structuring element for determining connectivity. If None, a
        cross-shaped structuring element is used.
    wrap_axes : tuple of int, optional
        Should labels be wrapped across array boundaries, and if so which axes.
        - (0,) only wrap over the 0th axis.
        - (0, 1) wrap over the 0th and 1st axis.
        - (0, 1, 3)  wrap over 0th, 1st and 3rd axis.
    trim_overlap : bool, optional
        If True, the overlap regions are trimmed from the output labels.
        Default is True.
    """
```

### Example

Consider the following problem.

The user wants to segment a large input image, i.e. a dask array.

```python
import numpy as np
import dask.array as da
import skimage
import matplotlib.pyplot as plt

dim = da.from_array(
    skimage.data.cells3d()[:, 1, :, :].max(axis=0), chunks=(110, 110))
```
image

and has a segmentation method that cannot be applied on the entire image (e.g. memory or method constraints) but works well on regions of the input image. The user can use `map_overlap` to apply the segmentation on the chunks of the input image and include overlap so that the segmentation method performs well on the chunk boundaries:

```python
model = models.Cellpose(gpu=False, model_type='cyto')

dseg = dim.map_overlap(
        lambda x: model.eval(x, diameter=None, channels=[0, 0])[0],
        dtype=np.uint16,
        depth=10,
    ).persist()

plt.figure()
plt.imshow(dseg)
```
image

At this point, the user faces two challenges:
1. the obtained segmentation labels show boundary artefacts and
3. they are not unique.

This is where `ndmeasure.merge_labels_across_chunk_boundaries` comes in to merge labels across chunk boundaries.

#### First use case / usage pattern
Similarly to the implementation of `ndmeasure.label` and e.g. the approach provided in [distributed cellpose](https://github.com/MouseLand/cellpose/pull/1062) (ping @GFleishman), one approach for merging labels consists in merging labels that touch across chunk boundaries. `ndmeasure.merge_labels_across_chunk_boundaries` allows to do this in a way that's agnostic of the segmentation method:

```python
import dask_image.ndmeasure

dseg_merged = dask_image.ndmeasure.merge_labels_across_chunk_boundaries(
    dseg
)['labels']

plt.figure()
plt.imshow(dseg_merged)
```

image

#### Second use case

The approach above joins all labels that form part of the same connected component in a boundary slice of thickness one. This can lead to artefacts. In fact, the example image above contains two merged cells which are clearly separated in the chunkwise segmentation. To separate these, `dask_image.ndmeasure.merge_labels_across_chunk_boundaries` has the option to merge labels based on the IoU within an overlap region.

```python
dseg = da.overlap.overlap(
        dim, depth={0: 10, 1: 10},
        boundary='none'
    ).map_blocks(
        lambda x: model.eval(x, diameter=None, channels=[0, 0])[0],
        dtype=np.uint16
    ).persist()

dseg_merged = dask_image.ndmeasure.merge_labels_across_chunk_boundaries(
    dseg, overlap_depth=10, iou_threshold=0.8,
)['labels'].compute(scheduler='single-threaded')

plt.figure()
plt.imshow(dseg_merged)
```
image

The labels of the two cells below which previously had been joined now remain separated.

## Finally

In my personal projects it's been useful to have a function that merges labels across chunks. I think `dask-image` could be a good place for providing such functionality to the community and I've had some offline feedback at meetings that this could be useful. Happy about any further feedback / discussion over here!

### DIFF
--- a/dask_image/ndmeasure/__init__.py
+++ b/dask_image/ndmeasure/__init__.py
@@ -360,8 +360,9 @@ def label(
         - (0, 1) wrap across the 0th and 1st axis.
         - (0, 1, 3)  wrap across 0th, 1st and 3rd axis.
     produce_sequential_labels : bool, optional
-        If False, the returned labels are not contiguous (i.e. 0, 1, 6045, 54654).
-        This is the default behavior and has advantages for parallelism.
+        If False, the returned labels are not contiguous
+        (i.e. 0, 1, 6045, 54654). This is the default behavior
+        and has advantages for parallelism.
         If True, the labels are relabeled to be sequential
         and contiguous from 1 to the number of features.
 
@@ -388,20 +389,21 @@ def label(
         wrap_axes=wrap_axes,
         produce_sequential_labels=False,
         )
-    
+
     relabeled = merged_labels_dict['labels']
     mapping = merged_labels_dict['mapping']
 
     if produce_sequential_labels:
         relabeled = relabel_sequential(relabeled)
         n = relabeled.max()
-        
+
     else:
         def count_labels(label_block):
             return len(np.unique(label_block[label_block > 0]))
 
         total = block_labeled.map_blocks(
-            lambda x: np.ones((1, ) * x.ndim, dtype=np.uint64) * count_labels(x),
+            lambda x: np.ones(
+                (1, ) * x.ndim, dtype=np.uint64) * count_labels(x),
             dtype=np.uint64,
             chunks=(1, ) * block_labeled.ndim
         ).sum()
@@ -432,7 +434,8 @@ def merge_labels_across_chunk_boundaries(
     chunk boundaries using a strategy dependent on ``overlap``:
     - If ``overlap > 0``, the overlap region between chunks is used to
       determine which between each pair of chunks should be merged.
-    - If ``overlap == 0``, labels that touch across chunk boundaries are merged.
+    - If ``overlap == 0``, labels that touch across chunk boundaries
+    are merged.
 
     Parameters
     ----------
@@ -444,8 +447,8 @@ def merge_labels_across_chunk_boundaries(
     iou_threshold : float, optional
         If ``overlap > 0``, the intersection-over-union (IoU) between labels
         in the overlap region is used to determine which labels should be
-        merged. If the IoU between two labels is greater than ``iou_threshold``,
-        they are merged. Default is 0.8.
+        merged. If the IoU between two labels is greater than
+        ``iou_threshold``, they are merged. Default is 0.8.
     structure : array of bool, optional
         Structuring element for determining connectivity. If None, a
         cross-shaped structuring element is used.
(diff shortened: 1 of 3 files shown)


m-albert:
Thanks @jni 🤗 Glad you like it and that you had the same idea!

Regarding this PR I still need to get an (obscurely) failing test under control...

Issue #428, opened 2026-04-22
Is dependency on dask[dataframe] really needed?


astrofrog:
For reproject we now depend on dask-image to use ``map_coordinates`` – however, because dask-image pulls in the dataframe-related optional extras from dask, our installation footprint has grown a lot:

https://github.com/astropy/reproject/issues/595

Having pyarrow and pandas (neither of which we need) is causing an installation of reproject to now take 400MB instead of 200MB.

I'm curious whether the dataframe extras is really needed as a required dependency here, or if dask-image could have a dataframe extras similar to dask if needed?

m-albert:
Hm `dask_image.ndinterp` depends on `dask.dataframe` for constructing output tables in `find_objects`:

https://github.com/dask/dask-image/blob/a513d9c32bfc1a6c271ba32410652e84ed20795c/dask_image/ndmeasure/__init__.py#L257-L259

That should be the only use of dask dataframe in the codebase. I agree that it's is a relatively large dependency and I think it makes sense to have it optional. What do you think @jakirkham @GenevieveBuckley ?

Cadair:
I noticed there's an explicit pandas dependency for the same (related?) bit of code, so that should be made optional at the same time if a change is made.

GenevieveBuckley:
I agree they could be made optional, provided ImportError handling is added to the find_objects function to inform users that it's required for that specific function.

m-albert:
Thanks @GenevieveBuckley 🙏

> provided ImportError handling is added to the find_objects function to inform users that it's required for that specific function.

I agree, that sounds like the right way here!

> I noticed there's an explicit pandas dependency for the same (related?) bit of code, so that should be made optional at the same time if a change is made.

That's a very good point @Cadair, `pandas` is also exclusively used in `find_objects` and can also be made optional in the same way.

@astrofrog @Cadair Feel free to open a PR about this!

astrofrog:
See https://github.com/dask/dask-image/pull/429

astrofrog:
Closed by https://github.com/dask/dask-image/pull/429

Issue #427, opened 2026-04-15
Compatibility with numpy v2


GenevieveBuckley:
Context: [PR 425](https://github.com/dask/dask-image/pull/425)
@avalentino says...

Actually there are 4 more tests failing when I run the test suite against numpy 2.4.
I was planning to fix also those ones but I dint have time so far.
I will open a new PR if I manage for find a fix.

The full log is at: https://ci.debian.net/packages/d/dask-image/unstable/amd64/69506502/

Pease find below the relevant part:
```
277s =================================== FAILURES ===================================
277s _ test_generic_filter_identity[<lambda>-1-None-generic_filter-generic_filter] __
277s 
277s sp_func = <function generic_filter at 0x7f893219c680>
277s da_func = <function generic_filter at 0x7f89321cd440>
277s function = <function <lambda> at 0x7f8931f9ae80>, size = 1, footprint = None
277s 
277s     @pytest.mark.parametrize(
277s         "sp_func, da_func",
277s         [(scipy.ndimage.generic_filter, dask_image.ndfilters.generic_filter)],
277s     )
277s     @pytest.mark.parametrize(
277s         "function, size, footprint",
277s         [
277s             (lambda x: x, 1, None),
277s             (lambda x: x, (1, 1), None),
277s             (lambda x: x, None, np.ones((1, 1))),
277s         ],
277s     )
277s     def test_generic_filter_identity(sp_func, da_func, function, size, footprint):
277s         a = np.arange(140.0).reshape(10, 14)
277s         d = da.from_array(a, chunks=(5, 7))
277s     
277s >       da.utils.assert_eq(d, da_func(d, function, size=size, footprint=footprint))
277s 
277s tests/test_dask_image/test_ndfilters/test__generic.py:82: 
277s _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
277s /usr/lib/python3/dist-packages/dask/array/utils.py:328: in assert_eq
277s     b, bdt, b_meta, b_computed = _get_dt_meta_computed(
277s /usr/lib/python3/dist-packages/dask/array/utils.py:279: in _get_dt_meta_computed
277s     x = _check_chunks(x, check_ndim=check_ndim, scheduler=scheduler)
[...]

PR #429, opened 2026-04-24
Make dataframe dependencies optional

astrofrog:
As discussed in https://github.com/dask/dask-image/issues/428, this makes the dataframe dependencies optional. I added a CI job to make sure everything works without the optional dependencies

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -34,7 +34,7 @@ jobs:
         shell: bash -l {0}
         run: |
           conda activate dask-image-testenv
-          python -m pip install -e .
+          python -m pip install -e .[dataframe]
           conda list
 
       - name: Run tests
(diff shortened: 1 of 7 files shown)


m-albert:
Generally looks good to me! But there seems to be a problem in the absence of the two dependencies..

astrofrog:
The failure was because of a change in recent versions of scipy, I've fixed compatibility with these versions now (just an issue with the tests). In short, ``generic_filter`` is pickier now and requires true scalars and doesn't accept length-1 arrays. This wasn't picked up before because scipy is pinned in the CI for other jobs.

m-albert:
Nice, completely agree with the way you split out the dependencies.

Thanks a lot also for fixing the testing issues you found! This fixes the first 4 test issues highlighted in https://github.com/dask/dask-image/issues/427 as well. Since this is not directly related to this PR, I'll open a new PR to add a test `test-latest-all-extras` to cover the case of latest environment + `find_objects`.

Before merging, could you add a test here which makes sure the `ImportError` is correctly raised when dataframe is not available?

m-albert:
Thank you @astrofrog ! Looks good to me.

Merging now.

Will try to create a release in the next days.

astrofrog:
@m-albert - just to check, do you think you would be able to do a new release soon with these changes? Thanks!

m-albert:
@astrofrog Sorry for the delay here. Planning to cut the release tomorrow afternoon!

m-albert:
Quick update: Release candidate is out on pypi but there are some problems on the conda feedstock that need fixing.

astrofrog:
Thank you for the update!

